PR #1, opened 2021-10-10
implemented TaskToEvent, TaskToEventAuto, TaskToEventManual (done by Emily and Valerie)

emilychangg:


src/main/java/TaskToEvent.java
```diff
@@ -1,2 +1,5 @@
-package main.java;public interface TaskToEvent {
+package main.java;
+
+public interface TaskToEvent {
+    public Event createEventFromTask(Task task, Schedule schedule)
```
ckatsube, line 4: Dependency Inversion. Make sure we're not making abstract things rely on concrete things (this will depend on whether Task ends up being abstract or not)

src/main/java/TaskToEventAuto.java
```diff
@@ -0,0 +1,54 @@
...
+     *          depending on the available times
+     */
+    @Override
+    public Event createEventFromTask(Task task, Schedule schedule) {
+        // if we have different schedules then the available time we get from for example the school schedule might be scheduled time in the family schedule
+        // potentially add a tag to events that will store the "schedule"
+
+        List<LocalDateTime> suggestedTimes = new ArrayList<>();
+        LocalDateTime availableTime = schedule.getAvailableTime(suggestedTimes, task.getDurationNeeded());
+        boolean scheduled = suggestTimeToUser(availableTime);
```
ckatsube, line 22: maybe `confirmTimeWithUser`? "suggest" doesn't seem to imply that we want a response back (it could be purely informational to the user).

src/main/java/TaskToEventAuto.java
```diff
@@ -0,0 +1,54 @@
...
+     * prints the time to the user and receives user input
+     * @param time an available time for the event
+     * @return  a boolean indicating if the user agrees with the suggested time
+     */
+    private boolean suggestTimeToUser(LocalDateTime time) {
+        boolean scheduled;
+        Scanner scanner = new Scanner(System.in);
+        System.out.println("Suggested time: " + time);
+        System.out.println("Type 0 for yes, 1 for no");
+        int response = scanner.nextInt(); //TODO exception handling
```
ckatsube, line 46: https://youtu.be/2a_ytyt9sf8?t=1500

I'm not saying don't commit TODOs, but make sure that you keep track of the TODOs you write and handle them quickly. TODOs are really technical debt that we don't want to accumulate a lot of.

src/main/java/TaskToEventAuto.java
```diff
@@ -0,0 +1,54 @@
...
+public class TaskToEventAuto implements TaskToEvent {
+    /**
+     *
+     * @param task      the task that needs to be converted into an event
+     * @param schedule  the schedule that the event will be added into
+     * @return an Event scheduled at a suggested time,
+     *          depending on the available times
+     */
+    @Override
+    public Event createEventFromTask(Task task, Schedule schedule) {
```
jenci2114, line 16: As discussed, replace Schedule with Calendar

src/main/java/TaskToEventAuto.java
```diff
@@ -0,0 +1,54 @@
...
+        Event event = new Event(Task, availableTime);
+        return event;
+    }
+
+    /**
+     * prints the time to the user and receives user input
+     * @param time an available time for the event
+     * @return  a boolean indicating if the user agrees with the suggested time
+     */
+    private boolean suggestTimeToUser(LocalDateTime time) {
```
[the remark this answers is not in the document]
emilychangg: Would moving it into another class be better?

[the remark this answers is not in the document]
ckatsube: At some point yep. I don't know if you've had your 207 lec yet but we went over how `Save` is a use case yet directly writes into the file system (an I/O operation). This kind of functionality should be done handled by Controllers or Presenters, which would be another class entirely. For the Phase 0 skeleton it should be fine, though.

PR #2, opened 2021-10-11
Implemented Task and Event Class

JunruL:
Since some decisions have not been made during the meeting, I made the decisions to write the complete code. These decisions are temporary and the structure and functionality of the classes may change after next meeting. Here are some ideas that need to be discussed:
- `Task` stores basic information of a task (name, completed, subtasks). It has a tree structure. 
- `MainTask` is a high level class which we create before any sub task is added. `MainTask` servers as a root of the task tree.
- Based on the implementation of `Task` and `MainTask`, any subtask will only have name, completed status, and subtasks. Subtask could not store the deadline and time needed. So subtasks are more like a check list instead of lists of true tasks. (Need to discuss this in next meeting. Another option is to timeNeeded and deadline both optional and combine Task and MainTask. However, as was discussed  in the meeting, the root task should have timeNeeded variable for it changing into an event, so it should not be optional for root task)
- `Event` and `Task` has some common methods so maybe they could have the same parent class
- Separate one-time events and repeated events may make the code look better (see `line 11` to `line 20` in `Event` class) Or anyone has ideas of how to store the repeat time information?

### DIFF
--- a/src/main/java/Task.java
+++ b/src/main/java/Task.java
@@ -7,7 +7,8 @@ import java.util.List;
 
 /**
  * A Task stores the name of the task, whether it is completed,
- * and subtasks (optional)
+ * the time needed to complete the task (optional), the deadline
+ * of the class (optional), and subtasks (optional)
  */
 public class Task {
 
--- a/src/main/java/Event.java
+++ b/src/main/java/Event.java
@@ -6,6 +6,11 @@ import java.time.LocalTime;
 import java.util.HashSet;
 import java.util.Set;
 
+/**
+ * An Event stores the name of the event, the start time
+ * and end time of the event, tags, the corresponding task,
+ * and the dates of event.
+ */
 public class Event {
 
     private String eventName;


src/main/java/Main.java
```diff
@@ -0,0 +1,9 @@
+package main.java;
+
+import java.time.Duration;
+import java.time.LocalDateTime;
+
+public interface Main {
```
jenci2114, line 6: The name of this interface sounds a little ambiguous to me

src/main/java/Task.java
```diff
@@ -0,0 +1,70 @@
...
+        this.taskName = newName;
+    }
+
+    public void complete() {
+        this.completed = true;
+    }
+
+    public void undoComplete() {
+        this.completed = false;
+    }
```
jenci2114, line 31: I would suggest merging those methods with one setter for `completed`

JunruL: Ok. And I will also rename other setters.

src/main/java/Main.java
jenci2114: If the interface is only for one specific class, then I think we would not need it unless there would be other similar classes that also hold root tasks in the future

PR #3, opened 2021-10-12
Added Calendar Class and its CRC Card

ckatsube:
Added the Calendar entity class

Details
Immutable 
- Events are passed in via constructor
- Mutability would remove the necessity of a `CalendarBuilder`

Functionality
1) finding an available time
2) checking if a specific time is available

Additional Notes
- Don't worry about the GapFinder and how the Interface and its implementation is not its own class for now. For the skeleton, the gap finding heuristic is just an internal independent system that can just be later extracted into its own files for extendability.
- I'm considering taking out the two above functionalities and making a new `CalendarAnalyzer` class (use case class) with the two above functions (which which also take in the `GapFinder`). Makes more sense from an OCP standpoint, too, as more complex queries would require opening up this class and editing it again.


src/main/java/Calendar.java
```diff
@@ -0,0 +1,107 @@
...
+    /**
+     * @return the name of this calendar. The name has no functional purpose.
+     */
+    public String getName() {
+        return name;
+    }
+
+}
+
+interface GapFinder {
```
ckatsube, line 71: See PR description for more about this.

PR #14, opened 2021-10-15
Tests

halfxblood:
Completed tests for Calendar class, Event class, Task class and Todolist class

### DIFF
--- a/tests/main/java/CalendarTest.java
+++ b/tests/main/java/CalendarTest.java
@@ -14,22 +14,18 @@ import static org.junit.jupiter.api.Assertions.*;
 
 class CalendarTest {
     List<LocalDateTime> timesToIgnore;
-    LocalDateTime startDate;
-    LocalTime endTime;
     Duration taskDuration;
-    Task task;
-    Event event;
     Calendar calendar;
 
     @BeforeEach
     void setUp() {
         timesToIgnore = new ArrayList<>();
         timesToIgnore.add(LocalDateTime.of(2021, 10, 14, 12, 0, 0));
-        startDate = LocalDateTime.of(2021, 10, 14, 14, 0, 0);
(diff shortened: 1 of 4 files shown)


tests/main/java/CalendarTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class CalendarTest {
+    List<LocalDateTime> timesToIgnore;
+    LocalDateTime startDate;
+    LocalTime endTime;
+    Duration taskDuration;
+    Task task;
+    Event event;
```
ckatsube, line 21: only need `timesToIgnore`, `calendar`, and `taskDuration`. You can add the others back if you end up making tests for them.

tests/main/java/EventTest.java
```diff
@@ -0,0 +1,73 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class EventTest {
+    LocalDateTime startDate;
```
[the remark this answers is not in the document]
ckatsube: to be clear, it doesn't need to be an instance variable. Can just be part of the setup

tests/main/java/EventTest.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    @Test
+    void getStartTime() {
+        assertEquals(startDate.toLocalTime(), event.getStartTime());
+    }
+
+    @Test
+    void getEndTime() {
+        assertEquals(LocalTime.of(16, 0), event.getEndTime());
```
ckatsube, line 43: can replace this new LocalTime with the instance var `endTime`

PR #17, opened 2021-10-15
Created Controllers for the TaskToEvent subclasses

ckatsube:
Added the Controller classes (and the interfaces) for the TaskToEvent subclasses. I feel like there was a better way to go about it, so feel free to make suggestions.


valyippee:
> Unless you're saying that we should have the TaskToEvent subclasses have the MainController which delegates to this smaller one.

Oh no I didn't mean that. I was thinking that the MainController should contain an instance of every other controller since it is a facade, so I meant that it should contain a TaskToEventController instance too. 

> Since both the controller and use case are paired, where would they be created?

Currently, the TaskToEvent subclasses depend on the interface that the TaskToEventConsoleController implements and I think that is correct based on dependency inversion (inner layer depend on an interface of the outer layer). But the problem now is that we have to find somewhere to instantiate the controller and the subclasses and pass the controller into the subclasses right? It sounds like something to be done on the controller layer so do we create them in MainController?

PR #20, opened 2021-10-16
Scheduling data accessors

TheXanMan321:


src/main/java/EventAdder.java
```diff
@@ -48,6 +47,6 @@ public boolean addRepeatedEvents(String eventName, LocalDateTime startTime,
         todoList.addTask(task);
         Event event = new Event(task, startTime, endTime.toLocalTime());
         event.addRepeatedDates(dates);
-        return eventScheduler.addEvent(event);
+        return true;
```
JunruL, line 50: Does it check whether the event has conflict with calendar?

TheXanMan321: No, and at the moment EventScheduler does not have a method to check if repeated dates interfere with a calendar.

valyippee: Made a note for this for after phase 0

PR #28, opened 2021-10-18
Controllers

jenci2114:
Implemented #5 on ApplicationDriver

### DIFF
--- a/crcCards/use_case/TaskToEventAuto
+++ b/crcCards/use_case/TaskToEventAuto
@@ -1,11 +1,7 @@
-Class name: TaskToEventAuto
-Implements TaskToEvent
+Interface: TaskToEventAuto
 
 Responsibilities:
-    - Finds an available time in the Calendar using EventScheduler and confirms it with the user
-    - Converts the Task into an Event with the available time
+    - Finds an available time in the Calendar using EventScheduler
 
 Collaborators:
-    - Task
-    - Event
-    - EventScheduler
+NA
--- a/crcCards/use_case/TaskToEventManual
+++ b/crcCards/use_case/TaskToEventManual
@@ -1,11 +1,7 @@
-Class name: TaskToEventManual
-Implements TaskToEvent
+Interface: TaskToEventManual
 
 Responsibilities:
-    - Creates an Event from the Task with a specific time specified by the user
-    - Uses EventScheduler to check if the time is available
+    - Checks the availability of a specific time specified by the user to be turned into an Event
 
 Collaborators:
-    - Task
-    - Event
-    - EventScheduler
+NA
--- a/src/main/java/ApplicationDriver.java
+++ b/src/main/java/ApplicationDriver.java
@@ -103,7 +103,10 @@ public class ApplicationDriver {
                 }
                 break;
             case "6":
-                System.out.println("Manual-scheduling is not yet implemented"); // TODO implement this functionality
+                printTasks();
+                Task taskManual = chooseTask();
+                LocalDateTime userSuggestedTime = inputTime();
+                controller.checkUserSuggestedTime(taskManual, userSuggestedTime);
                 break;
             default:
                 break;
@@ -238,6 +241,19 @@ public class ApplicationDriver {
         } while (!taskNames.contains(chosen));
 
         return controller.getTaskByName(chosen);
+    }
+
+    /**
+     * Prompts the user to input a time
+     * @return the time the user inputted
+     */
+    private static LocalDateTime inputTime() {
+        String format = "yyyy/MM/dd-HH:mm";
+        DateTimeFormatter formatter = DateTimeFormatter.ofPattern(format);
+        Scanner scanner = new Scanner(System.in);
+        System.out.print("Input your desired time in (" + format + ") (24 hour time)");
+        String timeString = scanner.nextLine();
+        return LocalDateTime.parse(timeString, formatter);
     }
 
     public static void main(String[] args) {
--- a/src/main/java/ApplicationDriver.java
+++ b/src/main/java/ApplicationDriver.java
@@ -105,8 +105,22 @@ public class ApplicationDriver {
             case "6":
                 printTasks();
                 Task taskManual = chooseTask();
-                LocalDateTime userSuggestedTime = inputTime();
-                controller.checkUserSuggestedTime(taskManual, userSuggestedTime);
+                LocalDateTime userSuggestedTime;
+                boolean timeAvailable;
+
+                do {
+                    userSuggestedTime = inputTime();
+                    timeAvailable = controller.checkUserSuggestedTime(taskManual, userSuggestedTime);
+                } while (!timeAvailable);
+
+                success = controller.createEvent(taskManual.getTaskName(), userSuggestedTime.toLocalTime(),
+                        userSuggestedTime.toLocalTime().plus(taskManual.getTimeNeeded()), new HashSet<>(), userSuggestedTime.toLocalDate());
+
+                if (success) {
+                    System.out.println("Event created from task");
+                } else {
+                    System.out.println("Failed to create event from task");
+                }
                 break;
             default:
                 break;
--- a/src/main/java/ApplicationDriver.java
+++ b/src/main/java/ApplicationDriver.java
@@ -1,6 +1,7 @@
 package main.java;
 
 import main.java.controller.MainController;
+import main.java.entity.Task;
 
 import java.time.Duration;
 import java.time.LocalDate;
@@ -73,17 +74,7 @@ public class ApplicationDriver {
                 }
                 break;
             case "2":
-                List<HashMap<String, String>> allTasksData = controller.getTasks();
-                if (allTasksData.size() == 0) {
-                    System.out.println("No tasks have been created");
-                }
-                for (HashMap<String, String> taskData : allTasksData) {
-                    String output = "Task: " + taskData.get("name") + ", "
-                                    + "deadline = " + taskData.get("deadline") + ", "
-                                    + "subtasks = " + taskData.get("subtasks") + ", "
-                                    + "completed = " + taskData.get("completed");
-                    System.out.println(output);
-                }
+                printTasks();
                 break;
             case "3":
                 boolean success = handleCreateTask();
@@ -102,7 +93,9 @@ public class ApplicationDriver {
                 }
                 break;
             case "5":
-                System.out.println("Auto-scheduling is not yet implemented"); // TODO implement this functionality
+                printTasks();
+                Task task = chooseTask();
+                success = controller.suggestTimeToUser(task);
                 break;
             case "6":
                 System.out.println("Manual-scheduling is not yet implemented"); // TODO implement this functionality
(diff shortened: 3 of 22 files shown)


src/main/java/ApplicationDriver.java
```diff
@@ -102,10 +93,34 @@ private static boolean handleQueryInput(String input) {
...
             case "6":
-                System.out.println("Manual-scheduling is not yet implemented"); // TODO implement this functionality
+                printTasks();
+                Task taskManual = chooseTask();
+                LocalDateTime userSuggestedTime;
+                boolean timeAvailable;
+
+                do {
+                    userSuggestedTime = inputTime();
+                    timeAvailable = controller.checkUserSuggestedTime(taskManual, userSuggestedTime);
```
ckatsube, line 113: Might be nice to have a message to tell the user why they need to resubmit.

src/main/java/controller/TaskToEventController.java
```diff
@@ -0,0 +1,57 @@
...
+                    task, eventController.calendarData.getCalendar(), eventController.eventScheduler, unwantedTimes
+            );
+            System.out.println("Suggested time: " + suggestedTime);
+            System.out.print("Type 'y' for yes, anything else for no: ");
+            response = scanner.nextLine();
+            // will be effective the next time the loop iterates
+            unwantedTimes.add(suggestedTime);
+        } while (!"y".equals(response));
+
+        return eventController.createEvent(task.getTaskName(), suggestedTime, task.getTimeNeeded());
```
ckatsube, line 44: Isn't this mutation? We want this method to just be a query. The controller should use the eventScheduler to create an event dependent upon this returned value (this is a concept called command/query separation)

src/main/java/entity/Task.java
```diff
@@ -24,7 +24,7 @@
      * @param taskName name of the task
      */
     public Task(String taskName) {
-        this(taskName, null, null);
+        this(taskName, Duration.ofHours(1), null);
```
ckatsube, line 27: I think the default duration should be a constant instead of an in-method literal.

src/main/java/use_case/EventScheduler.java
```diff
@@ -152,14 +153,16 @@ public int compareTo(TimeFrame o) {
...
     public LocalDateTime findTimeGap(List<TimeFrame> timeFramesToIgnore, Duration taskDuration) {
         Collections.sort(timeFramesToIgnore);
         TimeFrame first, second = null;
         for (TimeFrame t : timeFramesToIgnore) {
             first = second;
             second = t;
-            if (first != null && first.end.plus(taskDuration).isBefore(second.start))
+            if (first != null && first.end.plus(taskDuration).isBefore(second.start)) {
                 return first.end;
+            }
```
ckatsube, line 165: reversely, don't add parenthesis on single line if statements. We're going by Clean Architecture and Uncle Bob is on a mission to rid the world of curly braces.

src/main/java/ApplicationDriver.java
```diff
@@ -188,6 +203,58 @@ private static boolean handleCreateTask() {
...
+        List<String> taskNames = new ArrayList<>();
+        for (HashMap<String, String> task : allTasks) {
+            taskNames.add(task.get("name"));
+        }
+
+        Scanner scanner = new Scanner(System.in);
+        String chosen;
+
+        do {
+            System.out.print("Please choose a task by typing its name (case-sensitive): ");
```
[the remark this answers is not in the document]
jenci2114: I'll add a note here so that it may be changed in the future

src/main/java/use_case/EventScheduler.java
```diff
@@ -3,6 +3,7 @@
 import main.java.entity.Calendar;
 import main.java.entity.Event;
 import main.java.entity.Task;
+import org.jetbrains.annotations.NotNull;
```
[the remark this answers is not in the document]
jenci2114: Nope, I'll delete

crcCards/use_case/TaskToEventAuto
```diff
@@ -1,11 +1,7 @@
-Class name: TaskToEventAuto
-Implements TaskToEvent
+Interface: TaskToEventAuto
 
 Responsibilities:
-    - Finds an available time in the Calendar using EventScheduler and confirms it with the user
-    - Converts the Task into an Event with the available time
+    - Finds an available time in the Calendar using EventScheduler
```
JunruL, line 4: So now it only check the time and is not responsible for creating an event? If so, the name of the interface might not be proper

crcCards/use_case/TaskToEventManual
```diff
@@ -1,11 +1,7 @@
-Class name: TaskToEventManual
-Implements TaskToEvent
+Interface: TaskToEventManual
 
 Responsibilities:
-    - Creates an Event from the Task with a specific time specified by the user
-    - Uses EventScheduler to check if the time is available
+    - Checks the availability of a specific time specified by the user to be turned into an Event
```
JunruL, line 4: Same with the comment on `TaskToEventAuto`

src/main/java/use_case/EventScheduler.java
jenci2114: the code styling plugin i used told me to put curly braces around - I'll change it back

src/main/java/ApplicationDriver.java
```diff
@@ -102,10 +93,34 @@ private static boolean handleQueryInput(String input) {
...
                 break;
             case "5":
-                System.out.println("Auto-scheduling is not yet implemented"); // TODO implement this functionality
+                printTasks();
+                Task task = chooseTask();
+                success = controller.suggestTimeToUser(task);
+                if (success) {
+                    System.out.println("Event created from task");
+                } else {
+                    System.out.println("Failed to create event from task");
```
JunruL, line 102: It does not give information on why fail to create the event
Maybe use exception in the future so that it can deal with more cases?

src/main/java/ApplicationDriver.java
```diff
@@ -102,10 +93,34 @@ private static boolean handleQueryInput(String input) {
...
-                System.out.println("Manual-scheduling is not yet implemented"); // TODO implement this functionality
+                printTasks();
+                Task taskManual = chooseTask();
+                LocalDateTime userSuggestedTime;
+                boolean timeAvailable;
+
+                do {
+                    userSuggestedTime = inputTime();
+                    timeAvailable = controller.checkUserSuggestedTime(taskManual, userSuggestedTime);
+                } while (!timeAvailable);
```
JunruL, line 117: This block seems not parallel to the auto-scheduling, which calls a method instead using while loop here

src/main/java/controller/EventController.java
```diff
@@ -33,6 +34,24 @@
     /**
      * checks whether the time period is available to schedule a new event
      * and add the event if it is available
+     * @param eventName name of event
+     * @param startDateTime start time including date
+     * @param duration duration of event
+     * @return whether the creation succeeded
+     */
+    public boolean createEvent(String eventName, LocalDateTime startDateTime, Duration duration) {
+        LocalTime endTime = startDateTime.plus(duration.toSeconds(), ChronoUnit.SECONDS).toLocalTime();
```
JunruL: May exceed 24h

src/main/java/controller/TaskController.java
```diff
@@ -33,4 +34,12 @@ public boolean createTask(String taskName, Duration timeNeeded,
         return true; // TODO: return value should indicate success of data creation
     }
 
+    /**
+     * Gets a Task by its name
+     * @param name name of Task
+     * @return Task with given name
+     */
+    public Task getTaskByName(String name) {
```
JunruL, line 42: Can use the combination of name and start time as id, which will be unique.

src/main/java/controller/TaskToEventController.java
```diff
@@ -0,0 +1,57 @@
...
+    private final TaskToEvent taskToEvent = new TaskToEvent();
+
+    protected final EventController eventController;
+
+    public TaskToEventController(EventController eventController) {
+        this.eventController = eventController;
+    }
+
+    /**
+     * Suggest a time to the user until the user is agrees with the time
```
JunruL, line 21: grammar

src/main/java/controller/TaskToEventController.java
```diff
@@ -0,0 +1,57 @@
...
+
+        String response;
+        LocalDateTime suggestedTime;
+        Scanner scanner = new Scanner(System.in);
+
+        do {
+            suggestedTime = taskToEvent.getAvailableTime(
+                    task, eventController.calendarData.getCalendar(), eventController.eventScheduler, unwantedTimes
+            );
+            System.out.println("Suggested time: " + suggestedTime);
```
JunruL, line 37: So `TaskToEventController` will also interact with user? Not sure whether such design is ok

src/main/java/use_case/EventScheduler.java
```diff
@@ -112,7 +113,7 @@ public boolean checkAvailability(LocalDateTime targetTime, Calendar calendar, Du
                 if (targetTime.isAfter(startTime) && targetTime.isBefore(endTime)) {
                     return false;
                 }
-                else if (targetTime.plus(timeNeeded).isAfter(startTime)) {
+                else if (targetTime.plus(timeNeeded).isAfter(startTime) && targetTime.plus(timeNeeded).isBefore(endTime)) {
```
JunruL, line 115: another case is the target time duration covers an existing event

src/main/java/use_case/TaskGetter.java
```diff
@@ -42,4 +42,14 @@
 
         return task_data;
     }
+
+    public Task getTaskByName(String name, AccessTodoData accessTodoData) {
```
JunruL, line 46: Are AccessTodoData and TaskGetter in the same layer (or say, they will be used by same layer)?

src/main/java/use_case/TaskGetter.java
```diff
@@ -42,4 +42,14 @@
...
+        List<Task> allTasks = accessTodoData.getTodoList().getTasks();
+        for (Task task : allTasks) {
+            if (task.getTaskName().equals(name)) {
+                return task;
```
JunruL, line 50: Seems that `name` is used as id everywhere. So for now, when the user wants to create a new task or event, he will not allowed to create a repeated name. This probably requires less work to fix.

src/main/java/ApplicationDriver.java
jenci2114: ok i added a todo for future reference

src/main/java/ApplicationDriver.java
jenci2114: right, we could put the loop part in a controller class

src/main/java/controller/EventController.java
jenci2114: added note, we might need to use exceptions to achieve this

src/main/java/controller/TaskToEventController.java
jenci2114: lol fixed

src/main/java/use_case/TaskGetter.java
JunruL: OK this does not need to be fixed for phase 0 so we can choose a better method to fix it later

src/main/java/use_case/EventScheduler.java
jenci2114: fixed

src/main/java/use_case/TaskGetter.java
jenci2114: yeah they're both in the use case layer

src/main/java/use_case/TaskGetter.java
jenci2114: this method is temporary. After we have implemented the snowflake id thing we could binary-search the tasks

src/main/java/controller/TaskToEventController.java
JunruL: Putting it in an inner controller may force user to confirm a time. If we want to expand functionalities such as "cancel" (cancel creating event from this task), we may want to add this to the outer controller.

src/main/java/use_case/EventScheduler.java
```diff
@@ -111,8 +111,9 @@ public boolean checkAvailability(LocalDateTime targetTime, Calendar calendar, Du
                 LocalDateTime endTime = evt.getEndTime().atDate(date);
                 if (targetTime.isAfter(startTime) && targetTime.isBefore(endTime)) {
                     return false;
-                }
-                else if (targetTime.plus(timeNeeded).isAfter(startTime) && targetTime.plus(timeNeeded).isBefore(endTime)) {
+                } else if (targetTime.plus(timeNeeded).isAfter(startTime) && targetTime.plus(timeNeeded).isBefore(endTime)) {
+                    return false;
+                } else if (targetTime.plus(timeNeeded.dividedBy(2)).isAfter(startTime) && targetTime.plus(timeNeeded.dividedBy(2)).isBefore(endTime)) {
```
JunruL, line 116: seem to have bug
target time: 7:00 - 9:00
existing event: 8:58 - 8:59
BTW, I think it will be easier to check the cases when return true: target_end =< event_start or target_start >= event_end

jenci2114: we need to check for all current events so we can't early return true

added two more else if checks

PR #52, opened 2021-11-08
Renaming DTOs

ckatsube:
to be merged into #36 

(make sure to Squash when merging)

Renamed DTOs to make its purpose explicit as its usage is ephemeral and niche which makes it easy to question.


ckatsube:
Accidentally included some import optimization. Extracted that into #53 and will update this branch after that gets merged.

ckatsube:
closed for same reason as #53

PR #54, opened 2021-11-09
Notification System

JunruL:
Notification System [UML](https://drive.google.com/file/d/1d6-EMS59UJDCOAwQ8ZjQiKCUdbTaeNX0/view?usp=sharing)


ckatsube:
I've made the unfortunate mistake of not telling everyone to make PRs on `decouple-project`. Since I squashed that branch, I don't think we can actually merge this branch in. Can you create a new branch which you rebase onto main while deleting all of the squashed commits and then open a new PR for the rebased branch?
It'd look something like `git checkout notif-system` -> `git checkout -B notif-system-rebased` -> open Intellij's rebasing tool and select `main` with the `--interactive` option. Drop all the commits that you didn't write (I'll show an example below) and then keep accepting Your changes while just keeping the same commit messages. I've tested it and it seemed to work, but just in case rebuild (you might have to delete the two Task-Test.java files to compile). I didn't want to rebase and create the new feature branch myself because it'll say like "Valerie and Junru made changes committed by Craig" which is kind of weird since I didn't do anything.

Once you've rebased and made the new PR we can start reviewing without worries

image

Issue #57, opened 2021-11-09
Rebase Notification System to main branch


JunruL:
I've made the unfortunate mistake of not telling everyone to make PRs on `decouple-project`. Since I squashed that branch, I don't think we can actually merge this branch in. Can you create a new branch which you rebase onto main while deleting all of the squashed commits and then open a new PR for the rebased branch?
It'd look something like `git checkout notif-system` -> `git checkout -B notif-system-rebased` -> open Intellij's rebasing tool and select `main` with the `--interactive` option. Drop all the commits that you didn't write (I'll show an example below) and then keep accepting Your changes while just keeping the same commit messages. I've tested it and it seemed to work, but just in case rebuild (you might have to delete the two Task-Test.java files to compile). I didn't want to rebase and create the new feature branch myself because it'll say like "Valerie and Junru made changes committed by Craig" which is kind of weird since I didn't do anything.

Once you've rebased and made the new PR we can start reviewing without worries

![image](https://user-images.githubusercontent.com/13248104/140898667-14a716f0-d39b-4ea2-89d7-7b7d47f4f575.png)

_Originally posted by @ckatsube in https://github.com/CSC207-UofT/course-project-time/issues/54#issuecomment-963972214_

ckatsube:
Resolved in #58

PR #58, opened 2021-11-09
Notif system rebased

JunruL:
### **Issue #57** 
Close pull request #54 and rebase it to the `main` branch.

Notification System [UML](https://drive.google.com/file/d/1d6-EMS59UJDCOAwQ8ZjQiKCUdbTaeNX0/view?usp=sharing)


ckatsube:
### On choice of decorator pattern
Why was the decorator pattern (with the Notifier interface), specifically, chosen to implement notifying an arbitrary number and type of views?
How about just having a single iterable of Notifiers?
Or possibly having concrete NotificationObservers implemented in the adapter layer, which delegates the observed received notification (changed to a notification DTO) to the data converter (the true Presenter) which passes the converted data to views? Essentially the UI would create and subscribe concrete NotificationObservers hooked up to Presenters->Views to the system.

### On using enum
So enums are typically the bane of OCP because it requires you to add a new value when you want to add a new type, and then add a new case/switch or if branch for that new type everywhere the enum is used. The lower a policy is, the more we should design it to be resistant to change. In other words, it would be good to think about how we can take the enum out of Notification, our lowest level policy. The typical approach to pushing an enum upward in the policy layers is to take advantage of polymorphism. Some higher level module will eventually have to use if statements to actually generate the concrete classes, but it's much easier to change higher level modules.
A possible approach (granted I only thought of this for a few minutes) is to write a [Simple Factory (refactoring guru)](https://refactoring.guru/design-patterns/factory-comparison) which the NotificationController or Adder use to generate the respective Notification type. Also take note that you don't necessarily have to have a Notification interface and concrete Notification types, but you can have an attribute in the Notification entity that applies strategy pattern to the part that relates to behavior that may change across types.

### On CreationBoundaries
Unsure if this was made in relation to what was written in #36, but there are some mismatched relationships here.
An InputBoundary should the interface which serves as a clean single source of truth for the controller to use the application. Reworded, a controller should be using Boundary interfaces: the concrete boundaries should be invisible to the controller. It seems like the `NotificationCreationBoundary` is being used as a DTO to pass into the `NotificationAdder`, and should be named as so (NotificationCreationDTO or something in consistency with #52 which will be reopened later), and `NotificationAdder` should implement an interface which the controller uses instead (applying DIP on a higher to lower level module dependency). The Remover should also have its dependency inverted in the same manner, though imo the method can be changed to only use the id instead of the entire DTO.

src/main/java/entity/Task.java
```diff
@@ -4,21 +4,28 @@
...
  * the time needed to complete the task (optional), the deadline
  * of the class (optional), and subtasks (optional)
+ *
+ * - timeNeeded: the time duration needed to complete the task
+ * - notificationTimes: a set of times in Duration presenting how long in advance
+ *                      to send the event remainder notification
  */
 public class Task {
 
+    private final int id;
```
ckatsube, line 22: In the future try avoiding making small changes like this that aren't functionally relevant to the PR. Small refactoring changes should be put into small refactoring PRs

src/main/java/use_case/notification/NotificationCreationBoundary.java
```diff
@@ -0,0 +1,35 @@
+package main.java.use_case.notification;
+
+import main.java.constants.NotificationType;
+
+import java.time.Duration;
+
+public class NotificationCreationBoundary {
```
jenci2114, line 7: What is the purpose of this class? This class is not used by the controllers

src/main/java/use_case/notification/NotificationObserver.java
```diff
@@ -0,0 +1,117 @@
...
+     *
+     * @param taskId the id of the task
+     * @return string of message text for the task
+     */
+    private String taskMessage(int taskId) {
+        TaskReader taskReader = todoListManager.getTask(taskId);
+        String taskName = taskReader.getName();
+        LocalDate dueDate = taskReader.getDeadline().toLocalDate();
+        LocalTime dueTime = taskReader.getDeadline().toLocalTime();
+        return "You have a task " + taskName + " due on " + dueDate + " at " + dueTime + ".";
```
jenci2114, line 72: Should the text be generated by a use case?

src/main/java/use_case/notification/NotificationObserver.java
```diff
@@ -0,0 +1,117 @@
...
+    private String eventMessage(int eventId) {
+        EventReader eventReader = calendarManager.getEvent(eventId);
+        String eventName = eventReader.getName();
+        LocalTime startTime = eventReader.getStartTime();
+        LocalTime endTime = eventReader.getEndTime();
+        Set<LocalDate> dates = eventReader.getDates();
+        Set<Duration> notificationTimes = eventReader.getNotificationTimes();
+        LocalDateTime currentDateTime = LocalDateTime.now();
+        LocalDate date = searchEventDate(notificationTimes, dates, currentDateTime);
+        return "You have an upcoming event " + eventName +
```
jenci2114, line 93: same as above

src/main/java/use_case/notification/NotificationCreationBoundary.java
ckatsube: It's being used as a DTO in the NotificationController, lines 52 and 63.
The name doesn't line up with the purpose, though. At least not by the convention previously set. See On CreationBoundaries section of my comment above

src/main/java/entity/Event.java
```diff
@@ -9,17 +9,24 @@
...
+ * and the dates of event, and the notification times of the event.
+ *
+ * - dates: a set of LocalDate representing the recurring dates
+ * - notificationTimes: a set of times in Duration presenting how long in advance
+ *                      to send the event remainder notification
  */
 public class Event {
 
+    // TODO: make id final
+    private int id;
```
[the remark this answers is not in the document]
valyippee: Ok I have made the changes

src/main/java/entity/Task.java
valyippee: Should I rollback the changes for this part to make this pr clearer?

src/main/java/entity/Event.java
```diff
@@ -9,17 +9,24 @@
...
- * An Event stores the name of the event, the start time
- * and end time of the event, tags, the corresponding task,
- * and the dates of event.
+ * An Event stores the unique id, the name of the event, the start time
+ * and end time of the event, tags, the associated task,
+ * and the dates of event, and the notification times of the event.
+ *
+ * - dates: a set of LocalDate representing the recurring dates
+ * - notificationTimes: a set of times in Duration presenting how long in advance
+ *                      to send the event remainder notification
```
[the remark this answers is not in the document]
valyippee: I am not sure why but I don't see lines 16 and 17 on my side so I'll resolve this for now

src/main/java/use_case/notification/NotificationObserver.java
valyippee: That is a good point oops. At first Junru and I were thinking that since NotificationObserver is the one that differentiates "event" and "task" notification, might as well make it create the message that is specific to the type of notification. 

Should we move it to a presenter class instead?

src/main/java/entity/Event.java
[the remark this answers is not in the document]
valyippee: I moved to down to the javadoc of a method

src/main/java/entity/Task.java
ckatsube: Ah no sorry you don't need to. It's just something to take into consideration moving forward because it makes merge conflicts easy to occur. E.g. I think this might create conflict with the Id PR because that is changing the type of this attribute to long.

valyippee: Ah ok sorry about that. We can wait for the id pr to get merged and I'll resolve any conflicts before merging this pr into main.

src/main/java/use_case/notification/INotificationObserver.java
```diff
@@ -0,0 +1,7 @@
+package main.java.use_case.notification;
+
+import main.java.entity.Notification;
+
+public interface INotificationObserver {
```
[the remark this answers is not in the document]
valyippee: Ok I have renamed it

src/main/java/use_case/notification/INotificationTracker.java
```diff
@@ -0,0 +1,9 @@
+package main.java.use_case.notification;
+
+import main.java.entity.Notification;
+
+public interface INotificationTracker {
```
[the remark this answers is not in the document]
valyippee: Renamed

src/main/java/interface_adapters/notification/NotificationController.java
```diff
@@ -0,0 +1,81 @@
...
+        this.notificationPresenter = notificationPresenter;
+    }
+
+    /**
+     * Starts tracking if there are notifications to be sent.
+     * Should only be called once at runtime.
+     */
+    public void startTrackingNotifications() {
+        Thread thread = new Thread(this.notificationTracker);
+        thread.start();
```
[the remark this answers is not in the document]
valyippee: I added the synchronized keyword and I think that achieves it

src/main/java/entity/Task.java
valyippee: Ok I'll leave it as this for now and resolve any conflicts in the future

src/main/java/interface_adapters/notification/NotificationController.java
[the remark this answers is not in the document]
valyippee: I made the method synchronized and it should do the job

src/main/java/use_case/notification/NotificationCreationBoundary.java
valyippee: I changed the code to be consistent with the DTOs are you have mentioned

PR #60, opened 2021-11-09
Invert EventController's dependency on EventGetter to the CalendarEventDisplayBoundary interface

ckatsube:
Applying DIP to the dependency of the higher level module `EventController` on the lower level module `EventGetter`. This inversion also got rid of returning methods, which were replaced with just presentation calls.


ckatsube:
part of the #62 cleanup

ckatsube:
Changing approach to applying DIP

Issue #62, opened 2021-11-09
Dependency Inversion Cleanup


ckatsube:
Current DIP violations
EventController -> EventGetter and CalendarEventData
EventGetter ->EventInfoFromReader
TaskController -> TaskGetter and TodoListTaskData
TaskGetter -> TaskInfoFromReader
TaskToEventController -> EventScheduler and EventFromTaskId

PR #63, opened 2021-11-10
Removed TaskToEvent class as it was acting as an unproductive Middle Man

ckatsube:
Part of #62


ckatsube:
Instead of just dropping the middle man I also decided to include an interface between the controllers and use case. The current interface is a bit unstable and is probably going to be subject to changes or removal later, but I didn't want to introduce another DIP violation for no reason.

JunruL (approved):

PR #64, opened 2021-11-11
Added serialization for todolist and calendar

TheXanMan321:
Adds the ability to save events/tasks, and load them at the start. Currently only supports one Calendar and Todolist. To use the serializer, you need GSON. You can download the file from https://search.maven.org/artifact/com.google.code.gson/gson/2.8.9/jar


ckatsube:
I like the current work, but could I bother you to update the approach so that file reading and writing is a detail to the application, not integrated? Something like having a Use Case Interface which defines behavior for loading data and another for exporting data. You can then have another use case which accepts these and uses them appropriately. The UI/Controllers would be responsible for creating these concrete load/export classes, which are the ones that define the actual file reading and writing behavior. This approach should also allow the filepath to be passed in via the constructor, making it so that the use case isn't aware of the concept of a "filepath" but rather a "datasource/datadump."
The example might be shortsighted (as only just came up with it), but if you feel that making the reading and writing a detail in general is too complex a task we can stick with what you have for now.

src/main/java/interface_adapters/EventController.java
```diff
@@ -47,7 +48,7 @@ public EventController(CalendarEventCreationBoundary eventAdder, EventScheduler
     public boolean createEvent(String eventName, LocalDateTime startDateTime, Duration duration) {
         // todo use exceptions to ensure that duration won't last until the next day
         LocalTime endTime = startDateTime.plus(duration.getSeconds(), ChronoUnit.SECONDS).toLocalTime();
-        return createEvent(eventName, startDateTime.toLocalTime(), endTime, new HashSet<String>(), startDateTime.toLocalDate());
+        return createEvent(eventName, startDateTime.toLocalTime(), endTime, new HashSet<>(), startDateTime.toLocalDate());
```
ckatsube, line 51: Not going to ask you to change this, but going to keep adding reminders to keep unrelated changes to a minimum. If it's a small refactoring change I can review it quickly and get it accepted faster than this big PR.

ckatsube: The same applies to the CalendarEventModel modifier changes and EventGetter dependency changes, which one of us will probably get merge conflict issues because I'm removing those methods entirely in the Event dependency inversion PR.

src/main/java/entity_gateway/DatabaseManager.java
```diff
@@ -0,0 +1,34 @@
...
+import main.java.use_case.CalendarEventData;
+import main.java.use_case.CalendarEventModel;
+
+import java.io.*;
+import java.lang.reflect.Type;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.util.List;
+
+public class DatabaseManager {
```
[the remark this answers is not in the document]
TheXanMan321: Your right, I removed it.

ckatsube (approved):
Great work! Just some really small last changes

src/main/java/console_app/event_adapters/EventController.java
```diff
@@ -74,4 +79,9 @@ public boolean createEvent(String eventName, LocalTime startTime, LocalTime endT
         return false;
     }
 
+    public void saveEvents(String filename) throws IOException {
+        this.eventSaver.SaveEventData(filename);
```
valyippee, line 83: A really small detail! Can you change SaveEventData to camelCase?

valyippee (approved):

Issue #59, opened 2021-11-09
Update ID management capabilities


ckatsube:
Event and Task have ids but don't have an alg to go with it.

valyippee:
isnt this resolved with jenci's pr on the snowflake algo?

ckatsube:
Oh yep. Lemme add a quick link

PR #65, opened 2021-11-11
Id

jenci2114:
Implemented the snowflake algorithm for generating unique ids


src/main/java/use_case/Snowflake.java
```diff
@@ -0,0 +1,95 @@
...
+    private final long maxWorkerId = ~(-1L << workerIdBits);
+    private final long maxDatacenterId = ~(-1L << datacenterIdBits);
+    private final long maxSequenceBits = ~(-1L << sequenceBits);
+
+    // bit shifting for each id
+    private final long workerIdShift = sequenceBits;
+    private final long dataCenterIdShift = sequenceBits + workerIdBits;
+    private final long timestampShift = sequenceBits + workerIdBits + datacenterIdBits;
+
+    private final long initialTimeStamp = 1288834974657L;
```
ckatsube, line 26: kind of small but maybe change the variable name to denote that this is the initial unix timestamp in milliseconds since unix is typically denoted in seconds.

src/main/java/use_case/Snowflake.java
```diff
@@ -0,0 +1,95 @@
...
+    public synchronized long nextId() {
+        long timestamp = getTimestamp();
+
+        // check if clock has not moved backwards
+        if (timestamp < lastTimestamp) {
+            throw new RuntimeException(String.format(
+                    "Clock moved backwards. Refusing to generate id for %d milliseconds", lastTimestamp - timestamp));
+        }
+
+        // check if last timestamp equals the current timestamp
```
ckatsube, line 68: irrelevant comment. The condition reads `lastTimestamp == timestamp` which says enough

ckatsube: Same goes for the comment on lines 70, 77, and 81.
Comments should say important things that syntax, variable names, and method names do not. 
The comment on line 70 doesn't seem quite right. When we +1 and use the bitwise AND, we're not really increasing the sequence (in fact I think the bitwise & can only *decrease* the value, but I'm not 100% sure on that). What's the purpose of this operation and why are we using the bitwise AND? The answer to these kinds of things should be what comments entail.
The comment on 72 is ok, though I'd rather the purpose be the *reason* why we need to wait rather than just largely summarizing what the code pretty much already says.

src/main/java/use_case/Snowflake.java
```diff
@@ -0,0 +1,95 @@
+package main.java.use_case;
+
+/**
+ * Revised from Twitter's snowflake algorithm for generating unique IDs
+ * Github: https://github.com/twitter-archive/snowflake
```
ckatsube, line 5: Maybe be more specific in the source
https://github.com/twitter-archive/snowflake/blob/b3f6a3c6ca8e1b6847baa6ff42bf72201e2c2231/src/main/scala/com/twitter/service/snowflake/IdWorker.scala

src/main/java/use_case/Snowflake.java
ckatsube: Oh also the line 16 comment is pretty redundant, too.

src/main/java/use_case/Snowflake.java
```diff
@@ -0,0 +1,95 @@
...
+
+    // bit shifting for each id
+    private final long workerIdShift = sequenceBits;
+    private final long dataCenterIdShift = sequenceBits + workerIdBits;
+    private final long timestampShift = sequenceBits + workerIdBits + datacenterIdBits;
+
+    private final long initialTimeStamp = 1288834974657L;
+    private long lastTimestamp = -1L;
+
+    public Snowflake(long workerId, long datacenterId, long sequence) {
```
ckatsube, line 29: It would be nice to get some documentation on these parameters. Can these values be arbitrary? Should they represent some specific information? Do they *have to* represent some specific information?

src/main/java/use_case/Snowflake.java
jenci2114: Fixed

src/main/java/use_case/Snowflake.java
jenci2114: Fixed

src/main/java/use_case/Snowflake.java
jenci2114: OK changed

src/main/java/use_case/Snowflake.java
jenci2114: Documentation added

valyippee: Sorry on which lines are the documentation added? I don't seem to see them

valyippee (approved):

ckatsube:
resolved #59

PR #69, opened 2021-11-13
reschedule classes and interfaces

halfxblood:
Clear LocalTime and LocalDateTime


src/main/java/entity_gateway/EventEntityManager.java
```diff
@@ -43,4 +43,15 @@ public boolean addEvent(CalendarEventModel eventData) {
         }
         return eventReaderList;
     }
+    @Override
+    public boolean update(Long eventId, LocalDateTime startTime, LocalDateTime endTime) {
+        for(Event event: this.eventList){
+            if (event.getEventName().equals(eventId)){
```
valyippee, line 49: Shouldn't we check for if `event.getEventId().equals(eventId)` instead of using event name?

src/main/java/use_case/Reschedule.java
```diff
@@ -0,0 +1,26 @@
+package main.java.use_case;
+
+import main.java.entity_gateway.CalendarManager;
+import java.time.LocalDateTime;
+import java.util.List;
+
+public class Reschedule implements RescheduleAuto{
```
valyippee, line 7: Having a noun as the class name would be better. Maybe something like `Rescheduler` and `AutoRescheduler`

src/main/java/interface_adapters/RescheduleAutoController.java
```diff
@@ -0,0 +1,7 @@
+package main.java.interface_adapters;
+
+import main.java.use_case.TaskInfo;
+
+public interface RescheduleAutoController {
+    boolean rescheduleTime(TaskInfo task);
```
valyippee, line 6: This method should take in `EventInfo` instead of `TaskInfo` since we are rescheduling an existing event

valyippee (commented):
Apart from having the functionality of auto rescheduling an event, can you also add one for manual rescheduling? Like let the user input a time.

ckatsube:
See #85

PR #70, opened 2021-11-13
Event getid

halfxblood:
added getId() methods


ckatsube:
A different PR ended up implementing this

PR #75, opened 2021-11-14
Complete Tasks and Events

emilychangg:


src/main/java/data_gateway/CalendarManager.java
```diff
@@ -10,6 +10,8 @@
 
     boolean addEvent(CalendarEventModel eventData);
 
+    boolean completeEvent(String eventName);
```
valyippee, line 13: change it to `markEventAsCompleted`?

valyippee: also, this should take in id instead of name because that is the unique identifier of events

src/main/java/services/event_creation/CalendarEventCreationBoundary.java
```diff
@@ -4,4 +4,5 @@
 
     boolean addEvent(CalendarEventModel eventData);
 
+    boolean completeEvent(String eventName);
```
valyippee, line 7: change to `markEventAsCompleted`?

src/main/java/data_gateway/CalendarManager.java
emilychangg: can we merge in tahseens getId pr cause I'll need that for this

JunruL (approved):

PR #85, opened 2021-11-15
Fixed reschedule

halfxblood:


ckatsube:
scraped concept of rescheduling as a use case (in favor making it a GUI concept and just using updating use cases)

PR #92, opened 2021-11-22
Test (jenci)

jenci2114:


src/test/java/EventSchedulerTest.java
```diff
@@ -0,0 +1,191 @@
...
+        events.add(reader1);
+        events.add(reader2);
+    }
+
+    @Test
+    public void checkAvailabilityTrue() {
+        boolean expected = true;
+        boolean actual = eventScheduler.checkAvailability(LocalDateTime.of(
+                2021, 11, 25, 14, 30), Duration.ofHours(1));
+        assertEquals(expected, actual);
```
ckatsube, line 61: Maybe use assertTrue and assertFalse so the reader doesn't have to go up and check what expected's value is

src/test/java/EventSchedulerTest.java
```diff
@@ -0,0 +1,191 @@
...
+        manager = new MockCalendarManager(events);
+        eventScheduler = new EventScheduler(manager);
+
+        Set<String> tags1 = new HashSet<>();
+        tags1.add("CSC207");
+        tags1.add("Quiz");
+
+        Set<LocalDate> dates1 = new HashSet<>();
+        dates1.add(LocalDate.of(2021, 11, 25));
+        dates1.add(LocalDate.of(2021, 11, 27));
```
ckatsube, line 36: would be nice to have these Date/DateTime/Time literals to be named instead. It's a little bit more work but it helps get a better and faster grasp at what the test is actually doing.

src/test/java/EventSchedulerTest.java
jenci2114: Ok changed

src/test/java/EventSchedulerTest.java
jenci2114: So should I rename the variables to reflect the actual dates it contains (e.g. nov25Nov27)?

ckatsube: Yep personally I think that would help make it more readable.
Ultimately this is for readability, so if adding variables like nov25 makes it feel less readable, feel free to not make the change.

src/test/java/TaskToEventTest.java
```diff
@@ -0,0 +1,157 @@
...
+        public void saveEvents(String savePath) throws IOException {
+
+        }
+    }
+
+    private class MockEventReader implements EventReader {
+
+        @Override
+        public long getId() {
+            return 0;
```
JunruL, line 124: Should be 0L?

src/test/java/TaskToEventTest.java
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        @Override
+        public String getName() {
+            return null;
+        }
+
+        @Override
+        public LocalTime getStartTime() {
```
JunruL, line 133: To test conflict, we should set some start and end time

src/test/java/TaskToEventTest.java
```diff
@@ -0,0 +1,157 @@
...
+    @Test
+    void getAvailableTime() {
+        LocalDateTime expected = scheduler.getAvailableTime(
+                new ArrayList<>() , taskInfo.getDuration()).truncatedTo(ChronoUnit.MINUTES);
+        LocalDateTime actual = taskToEvent.getAvailableTime(taskInfo, scheduler, new ArrayList<>());
+        assertEquals(expected, actual);
+    }
+
+    @Test
+    void checkTimeAvailability() {
```
JunruL, line 46: There should be several test cases:
Conflict does not exist
Conflict exists

src/test/java/TaskToEventTest.java
```diff
@@ -0,0 +1,157 @@
...
+        LocalDateTime userSuggestedTime = LocalDateTime.of(2021, 11, 26, 10, 0);
+        boolean expected = scheduler.isAvailable(userSuggestedTime.toLocalTime(), taskInfo.getDuration(),
```
JunruL, line 48: You should not be calling the method for expected value. It should a hard coded.

src/test/java/EventSchedulerTest.java
jenci2114: Got it

src/test/java/TaskToEventTest.java
jenci2114: I think either way is fine for small numbers

src/test/java/TaskToEventTest.java
jenci2114: Conflict checking is EventScheduler's job, and the conflict testings are in EventSchedulerTest (since the purposes of TaskToEvent involves calling EventScheduler)

src/test/java/TaskToEventTest.java
jenci2114: same as above

src/test/java/TaskToEventTest.java
jenci2114: Calling EventScheduler is the expected behaviour of this method (see Line 19 and Line 31 of TestToEvent)

src/test/java/TaskToEventTest.java
JunruL: OK

src/test/java/TaskToEventTest.java
JunruL: Oh I see. You are correct.

src/test/java/EventSchedulerTest.java
ckatsube: I see you've changed some of the literals. Did you choose to not do the rest for a specific reason? As an example, I see a lot of November 25 dates, so it might be nice to have a more global `nov25 = LocalDate.of(2021, 11, 25)` and then use `nov25.atTime(insert time)` for LocalDateTimes

PR #95, opened 2021-11-23
Adding Event Dates Generation Strategies

### LINKED ISSUE #51: Change Event Entity such that it accepts a "date strategy"
Currently, Event holds a Set of LocalDates, but it should be changed to allow generic repetition of dates, such as "repeat every month".

ckatsube:
Closes #51 

Because concepts like "every week" cannot be directly defined with a list of dates (as the concept extends infinitely), we have to reduce the problem to implicitly defining dates and creating the explicit dates when certain concrete conditions are met.

The specific approach in this PR takes date generation strategies and creates lists of concrete dates within a certain queried date range.
To allow for combinations and variations of strategies, the Composite Pattern (for combining) and Decorator Pattern (for restricting) were included, along with a few basic implementations of them.

Due to the complexity of combinations of strategies, it's reasonable to use a builder (or factory, though builder was chosen here) to abstract the creation process away from the concrete strategies.

========= at time of original draft
The current builder, however, is complex in its own right and ideally a BuilderDirector should be created to abstract even this process away into ideally a more declarative-type process.
I also imagine that the Director would accept a Form, which is able to be validated possibly via converting to a String and applying regex validation (e.g. a form creating a Weekly event on Monday at noon might be stringified as "W(Mon.12:00)")


ckatsube:
Since I completed the declarative building process, I just wanted to get these changes in.

A few things to note
1) the builder is buggy if you try using it directly. Using the DatesForm and Director should be OK.
2) I didn't update Events to actually utilize the strategies. Currently the strategies are being used to just recreate the previous single-date attributes and whatnot. This is largely because there were a lot of changes in the process that felt out of the scope of just introducing the strategies. 
3) I think you can ignore the actual building process and strategy structure. The only things that really matter are what things the builder/director make available as well as the interface of the strategy. 
4) Planning on introducing Duration to the Strategy's method (might be its own pr, might be part of when I try to add it to events)
5) In the future I plan on creating a repository class for the Strategies like a StrategyManager, but it might be named differently because I envision that the internal ids are session-specific and there can be a different form of the strategy that can be used to store/rebuild it (I already have an idea for this).
6) Yes DatesForm is basically just a name wrapper for the Iterator. It just made sense to me to have it since its like a part of the program
7) MultipleBasicForm is more akin to a VM/Presenter(MVP) or Controller (CA) class than a use case. It actually doesn't make a ton of sense for the console application, but I wrote it for the GUI ppl to use. If we want to better apply architectural boundaries to it, we can just make the DatesForm itself some functional class that rules can be added to, and the MultiBasicForm is just a pure controller(CA)/presenter(MVP)/VM which manipulates it.

src/main/java/console_app/ApplicationDriver.java
```diff
@@ -341,6 +343,60 @@ private static boolean isInt(String string) {
...
+            index++;
+
+        DayOfWeek day = DayOfWeek.of(index + 1);
+
+        System.out.println("When in the day do you want the event to reoccur (HH:mm)?");
+        Scanner scanner = new Scanner(System.in);
+        String timeString = scanner.nextLine();
+        LocalTime timeOfDay = LocalTime.parse(timeString);
+
+        System.out.println("(Optional) input the time to start from (enter nothing to not use)");
```
emilychangg, line 369: What's the difference between this and the one above (when in the day do you want the event to reoccur)?

emilychangg: or is this and the one below duration?

ckatsube: Aha good question. It's supposed to be when the Weekly occurrences should start from/end by. E.g. Every Monday at 13:00 from September 1 until December 31.

I can make the messages clearer.

emilychangg: ah I see. Yea clearer message would be good, I also think including a format thing like you did above would be helpful too. Like when you added (HH:mm) to the prompt above this one.

emilychangg:
Should we add a way to also add the date strategy when you create an event from the task?

src/main/java/services/event_creation/EventAdder.java
```diff
@@ -12,7 +20,18 @@ public EventAdder(CalendarManager calendarManager) {
...
-        calendarManager.addEvent(eventData);
+
+        String eventName = eventData.getName();
+        HashSet<String> tags = eventData.getTags();
+        DatesForm form = eventData.getForm();
+        Duration eventDuration = eventData.getDuration();
+
+        StrategyBuilderDirector director = new StrategyBuilderDirector();
+        DateStrategy strategy = director.createStrategy(form);
+        List<LocalDateTime> times = strategy.datesBetween(LocalDateTime.now(), LocalDateTime.now().plusYears(1));
```
emilychangg, line 31: does this list of times ever get used again?

emilychangg: Like is there currently a way to view all scheduled events from now until x date

emilychangg: oh nevermind I just reread your initial comment on this PR

emilychangg (approved):

Issue #90, opened 2021-11-21
Displaying Pomodoro timer seconds in UI


ckatsube:
Pomodoro timer currently only runs an action when it ends, so the current time on the clock is not visible unless calculated implicitly by the UI.

PR #96, opened 2021-11-25
Added Circles for Pomodoro

TheXanMan321:
Added a circle timer display on the Pomodoro page, which counts down and alternates between fixed work and break times (1 minute each). There isn't any other functionality on the page, like the ability to input work and break times, or a display of the actual time.


ckatsube:
related to (but not closing) #90

Issue #101, opened 2021-11-25
Move markEventAsCompleted method out of the event adding class


valyippee:
Currently, ```markEventAsCompleted()``` is in ```EventAdder```, but it should be moved to another class whose responsibility is to update existing events. Same for tasks.

ckatsube:
Linking instead of closing directly

PR #102, opened 2021-11-26
Update entities

halfxblood:


src/main/java/data_gateway/EventEntityManager.java
```diff
@@ -103,4 +104,33 @@ public void markEventAsCompleted(long eventId) {
         }
         return eventReaderList;
     }
+
+
+    @Override
+    public void updateName(long id, String newName) {
+        for(Event event : eventList){
```
ckatsube, line 111: Might be a good idea to add a private `getById` method to prevent this repetitive for loop (same for the TaskManager)

src/main/java/data_gateway/CalendarManager.java
```diff
@@ -1,8 +1,11 @@
 package data_gateway;
 
+import entity.Event;
```
ckatsube, line 3: Make sure to clean up imports (IntelliJ has an optimize imports option)

src/main/java/services/update_entities/UpdateEvent.java
```diff
@@ -0,0 +1,32 @@
...
+    public UpdateEvent(EventEntityManager eventEntityManager, long id){
+        this.calendarManager = eventEntityManager;
+        this.id = id;
+    }
+
+    public void updateName(String newName){
+        calendarManager.updateName(id, newName);
+    }
+
+    public void updateStartTime(LocalTime startTime){
```
ckatsube, line 21: Since these are actually not in use right now, how about adding a new option to the ApplicationDriver to show a basic usage of editing?

ckatsube:
Closes #101

src/main/java/data_gateway/CalendarManager.java
```diff
@@ -13,6 +16,12 @@
 
     List<EventReader> getAllEvents();
 
+    void updateName(long id, String newName);
+
+    void updateStartTime(long id, LocalTime newStartTime);
+
+    void updateEndTime(long id, LocalTime newEndTime);
+
```
valyippee, line 24: Can we also have an `updateTags()` method? I think you would have to change `tags` in the `Event` class to become not final too

src/main/java/data_gateway/TodoListManager.java
```diff
@@ -16,6 +19,12 @@
 
     void completeTask(long taskId);
 
+    void updateName(long id, String newName);
+
+    void updateDuration(long id, Duration newDuration);
+
+    void updateDeadline(long id, LocalDateTime newDeadline);
+
```
valyippee, line 27: Can we add in `updateSubTasks()` for `Task` too?

src/main/java/services/update_entities/UpdateEvent.java
```diff
@@ -0,0 +1,32 @@
+package services.update_entities;
+
+import data_gateway.CalendarManager;
+import data_gateway.EventEntityManager;
+
+import java.time.LocalTime;
+
+public class UpdateEvent {
```
valyippee, line 8: It seems that you are creating an `UpdateEvent` class whenever someone wants to update an entity. But it makes more sense to me that this class should have methods that take in `id` as input and make the updates from there, so there will be only one existence of this class in the whole program, like our other use case classes.

ckatsube (approved):

ckatsube:
closed in favor of #115

PR #103, opened 2021-11-27
change the return type to long

JunruL:
The return type of some adding methods are still `void` after PR #97. I updated them in this PR.


ckatsube:
Does the client/VM/Presenter actually need to be exposed to the created entity's id? I don't see the return being used in the ApplicationDriver or any of the controllers. What's the intended purpose/usage?

ckatsube (approved):

PR #104, opened 2021-11-27
Removed Auto Task-To-Event Converting

### LINKED ISSUE #61: Cleaning up TaskToEvent feature
Removing the `TaskToEvent` class middle man
Removing of the getting & choosing process to presenting and choosing instead (create acyclic dependency)

ckatsube:
Replaced auto behavior with just presenting the suggested time and having the user manually accept or choose their own time.

closes #61

### DIFF
--- a/src/main/java/console_app/ApplicationDriver.java
+++ b/src/main/java/console_app/ApplicationDriver.java
@@ -39,12 +39,11 @@ public class ApplicationDriver {
         queryMenu.put("2", "View all tasks");
         queryMenu.put("3", "Create a new task");
         queryMenu.put("4", "Create a new event");
-        queryMenu.put("5", "Auto schedule a task");
-        queryMenu.put("6", "Manually schedule a task");
-        queryMenu.put("7", "Mark a task as completed");
-        queryMenu.put("8", "Mark an event as completed");
-        queryMenu.put("9", "Save my Data");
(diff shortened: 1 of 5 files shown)


src/main/java/console_app/ApplicationDriver.java
```diff
@@ -93,50 +92,45 @@ private boolean handleQueryInput(String input) {
...
-                }
-                break;
-            case "6":
-                positionToIdMapping = controller.presentAllTasksForUserSelection();
-                if (positionToIdMapping.size() != 0) {
-                    TaskInfo taskManual = chooseTask(positionToIdMapping);
+                    LocalDateTime suggestedTime = controller.getSuggestedTime(taskInfo.getDuration());
+                    String formattedTime = suggestedTime.format(DateTimeFormatter.ofPattern("yyyy/MM/dd-HH:mm"));
+                    System.out.println("Suggested time is: " + formattedTime);
+
```
valyippee, line 98: Can you add in a check for whether the user agrees with the suggested input?

valyippee (approved):

src/main/java/console_app/ApplicationDriver.java
ckatsube: Great suggestion. I added a 'y' option.

PR #105, opened 2021-11-28
adding the actual timer to the pomodoro circle

emilychangg:
display the time left in the current pomodoro interval
added error checking so the user will not be able to input anything other than an integer
user can input their own work and break time intervals or use the default intervals


src/main/java/data_gateway/CalendarManager.java
```diff
@@ -7,7 +7,7 @@
 
 public interface CalendarManager {
 
-    void addEvent(CalendarEventModel eventData);
+    long addEvent(CalendarEventModel eventData);
```
TheXanMan321, line 10: Should the changes to Event, CalendarAnalyzer and Tasks be in another PR?

src/main/java/gui/ClockController.java
```diff
@@ -89,8 +96,11 @@ private void switchCircle(GraphicsContext gc) {
 
     @FXML
     void startClock(MouseEvent event) {
+        PomodoroRunner pomodoroRunner = new PomodoroRunner();
```
TheXanMan321, line 99: I think it would be easier to update the label in the AnimationTimer's loop, rather than opening another thread. `elapsedTime` reps the number of nanoseconds that has passed, you could pass it to the runner to format it.

TheXanMan321 (commented):
Did we decide to have the timer work in the GUI? If so, I think starting/stopping the timer and updating labels can be handled by an outer layer, and we won't need most of the files in pomodoro_running

src/main/java/data_gateway/CalendarManager.java
emilychangg: Yea this wasn't my work I don't know how this ended up in my pr and I'm not too sure how to remove it either. Also I think I can move the getCurrentTime function into the ClockController class and I'm looking into the task functionality of javafx right now to handle it

src/main/java/gui/ClockController.java
emilychangg: Sorry what's the runner?

emilychangg: Ah I see, I'll try that out

src/main/java/data_gateway/CalendarManager.java
ckatsube: might be because you haven't merged main into `pomodoro_gui` (the target branch)

ckatsube:
Great work, just a few comments.
In the GUI, it'd be nice to have a little warning and disabled start button instead of an error window for invalid input.
I legitimately don't think it's worth having the controller depend on the service classes like `TimeFormatter` if that's the only dependency. It's a bit sad to delete the Pomodoro service stuff, but it's still in the git history so if we ever end up coming back to needing it we'll be able to retrieve it fairly easily.

PR #106, opened 2021-11-29
Added entity and data gateway layer for notification system

valyippee:
Added a new entity: Notification, and the data gateway classes for notification and notification settings. Right now, notification settings store a default email address and have both email and desktop notifications enabled (did not persist them to the database). The `JsonNotificationAdaptor` is not completed as I am not familiar with persisting data to JSON file.

Repackaged the `data_gateway` folder too as there are too many classes.


src/main/java/data_gateway/notification/NotificationEntityManager.java
```diff
@@ -0,0 +1,121 @@
...
+        for (Notification notification : notifications) {
+            notificationReaderList.add(new NotificationToNotificationReader(notification));
+        }
+        return notificationReaderList;
+    }
+
+    /**
+     * Return a notification.
+     * @param associatedId the id of the associated object that the notification is for
+     * @param timeInAdvance the duration in advance that the notification was to be sent out
```
ckatsube, line 75: Should this be "was to be sent out" or "is to be sent out"?

src/main/java/data_gateway/notification/NotificationEntityManager.java
```diff
@@ -0,0 +1,121 @@
...
+     * @return the corresponding notification
+     */
+    @Override
+    public NotificationReader getNotification(long associatedId, Duration timeInAdvance) {
```
ckatsube, line 79: Why is a notification identified by both its id and time in advance? Is there a reason it can't just be identified by id?

Same question for delete

I guess a more specific question is, you're already injecting in the Snowflake, so why not generate ids for the notifs instead of using the associated id as its own id?

ckatsube: ^ after some external discussion we decided that notifications would only be accessed in the context of its associated object, hence they should not need their own ids for identification. An associated object may have multiple times, hence the two identifying pieces of data.

src/main/java/data_gateway/notification/NotificationEntityManager.java
valyippee: You are right it should be "is". Fixed that wording

ckatsube (approved):

PR #108, opened 2021-12-01
Added view for calendar and classes for navigation panel

valyippee:
This pr contains added classes for
1) views of the calendar (monthly, weekly, and daily page)
2) views of navigation panels (collapsed and extended)
3) a navigation helper class (to facilitate switching views)

These are just the views (no view models or models). You don't have to look at the fxml files. Just run the program and see if everything looks ok from there. 

I am more concerned about number 3, the use of a navigation helper class. I'm not sure if that is the best way to do it, but that was the way that I thought of extracting the navigation details from other classes. If you have any better ideas please let me know!


ckatsube (approved):
Two things stood out to me when messing around with the gui.
1) The nav bar consistently flickers in and out which is quite distracting. I think the miniaturized nav bar is already good, so it can just be static there while possibly expanding into the more detailed bar. Furthermore you can accidentally drag-flick the mini nav bar off the screen and end up never being able to get it back.
2) When you swap from Month to Week view, the name of the last week/first week is the latter month; however, when you start going left/right between the weeks and return to the last week/first week it's the former month (similarly hitting the Today buttons makes it the latter month). Not sure how much control you have over this but this should be consistent.

Regarding the Navigation Utility class.
I can definitely see its convenience, and for a small app I think it'll serve us fine. However I'd imagine that in an MVVM architecture, this utility class would really be a View which observes a VM that contains "meta-states" and that the other views trigger the VM to cycle through these meta-states. The View classes (primarily this Nav Util class) observes that the VM changed to its "calendar meta state" and hence it'll trigger the `enterMonthlyCalendarPage` or whatever corresponding JavaFX-rendering method is appropriate.
The `initializeNavPanel` seems to make the most sense as an actual static utility method, though in the sense that other concrete Views would want to render the nav panel as well.

Overall some great work. I'm hitting approve 1) for the sake of time 2) because I don't think the things I mentioned are worth holding this PR back, but it's good to think about moving forward.

PR #109, opened 2021-12-01
Application Factories

ckatsube:
Introduces factories for many of the major application components.

There are a lot of concrete dependencies in the program right now, which are reflected in the factories (TaskGetter, EventGetter, etc)

The -Presenter classes didn't get included in the factories primarily because I thought that the client code would want more control over picking specific implementations.


src/main/java/services/services_factory/BasicObservableRepositoryFactory.java
```diff
@@ -0,0 +1,36 @@
+package services.services_factory;
+
+import data_gateway.event.ObservableEventEntityManager;
+import data_gateway.event.ObservableEventRepository;
+import data_gateway.task.ObservableTaskEntityManager;
+import data_gateway.task.ObservableTaskRepository;
+
+public class BasicObservableRepositoryFactory implements ObservableRepositoryFactory {
```
jenci2114, line 8: What are the purposes of the basic factories? If they have specific purposes, maybe consider naming the classes more explicitly

ckatsube: The Basic- prefix was just to get around the issue of the interface having the same name. There's not really much else to say other than it's the observable repository factory. Maybe that it delegates to another factory? DecoratingObservableRepositoryFactory idk xD

jenci2114: I see

jenci2114 (approved):

valyippee (approved):

PR #110, opened 2021-12-01
Removed Concrete Dependency on EventGetter

ckatsube:
Segregated `EventGetter` into two classes, both implementing their own interfaces.
`EventGetter` implements a `-RequestBoundary` (used for `get-()` methods) and a new `EventOutputter` implements the old `-DisplayBoundary` (used for `present-()` methods)

related to #62


valyippee (approved):

src/main/java/console_app/event_adapters/EventController.java
```diff
@@ -16,12 +19,18 @@
 public class EventController {
 
     private final CalendarEventCreationBoundary eventAdder;
-    private final EventGetter eventGetter;
+    private final CalendarAnalyzer eventScheduler;
```
JunruL, line 22: It seems that eventScheduler is not used. Is it added for future function?

ckatsube: Seems like I accidentally added it when merging main. Thanks for the catch

PR #113, opened 2021-12-01
Introducing Observable Entity Repositories

ckatsube:
included applying them to the console app. The ConsoleApp implementation is a bit crude, using dynamically defined classes and lambda methods, but it was primarily just an example of a quickly define listener/observer 

The primary purpose of the observable repositories is for ease in implementing the MVVM UI architecture. Having the Repositories being observable means that the VM be more easily synced with our application domain.


valyippee (approved):

src/main/java/data_gateway/event/ObservableEventEntityManager.java
```diff
@@ -0,0 +1,70 @@
...
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.util.HashSet;
+import java.util.List;
+
+public class ObservableEventEntityManager implements ObservableEventManager {
+
+    private final CalendarManager calendarManager;
+    private final ObservableProperty<EventReader, Long> newEventObservers;
+    private final ObservableProperty<EventReader, Boolean> updateCompletionObservers;
```
valyippee, line 16: Do we have observers for when events are updated?

src/main/java/data_gateway/ObservableProperty.java
```diff
@@ -0,0 +1,20 @@
...
+public class ObservableProperty<E, T> {
+
+    private final List<PropertyObserver<E, T>> observers = new ArrayList<>();
+
+    public void addObserver(PropertyObserver<E, T> observer) {
+        observers.add(observer);
+    }
+
+    public void notifyObservers(E entity, T value) {
+        for (PropertyObserver<E, T> observer : observers)
```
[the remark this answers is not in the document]
ckatsube: No since it's a one-line loop.

src/main/java/data_gateway/task/ObservableTaskManager.java
```diff
@@ -0,0 +1,10 @@
+package data_gateway.task;
+
+import data_gateway.ObservableRepository;
+
+/**
+ * Interface allowing implementors polymorphic access to
+ * {@link TodoListManager} and {@link ObservableRepository<TaskReader>}
+ */
+public interface ObservableTaskManager extends ObservableRepository<TaskReader>, TodoListManager {
```
valyippee, line 9: Name of the interface isn't standardized with the one for event (`ObservableEventRepository`)

PR #121, opened 2021-12-02
Test jenci

jenci2114:
Added 4 tests

### DIFF
--- a/src/main/java/services/event_creation/EventInfoFromReader.java
+++ b/src/main/java/services/event_creation/EventInfoFromReader.java
@@ -51,16 +51,4 @@ public class EventInfoFromReader implements EventInfo {
         return eventReader.getCompleted();
     }
 
-    @Override
-    public boolean equals(Object o) {
-        if (this == o) return true;
-        if (o == null || getClass() != o.getClass()) return false;
-        EventInfoFromReader that = (EventInfoFromReader) o;
-        return Objects.equals(eventReader, that.eventReader);
-    }
-
-    @Override
-    public int hashCode() {
-        return Objects.hash(eventReader);
-    }
 }
--- a/src/main/java/services/event_creation/EventInfoFromReader.java
+++ b/src/main/java/services/event_creation/EventInfoFromReader.java
@@ -5,6 +5,7 @@ import services.event_presentation.EventInfo;
 
 import java.time.LocalDate;
 import java.time.LocalTime;
+import java.util.Objects;
 import java.util.Set;
 
 public class EventInfoFromReader implements EventInfo {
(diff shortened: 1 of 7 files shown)


src/test/java/EventGetterTest.java
```diff
@@ -0,0 +1,202 @@
...
+import services.event_creation.EventInfoFromReader;
+import services.event_presentation.CalendarEventPresenter;
+import services.event_presentation.EventGetter;
+import services.event_presentation.EventInfo;
+
+import java.io.IOException;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.LocalTime;
+import java.util.*;
```
ckatsube, line 16: Change your IDE settings to not bundle imports into the wildcard (you should be able to set the count to like 999)

src/test/java/TaskOutputterTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+        }
+
+        @Override
+        public void saveTodo(String filepath) throws IOException {
+
+        }
+    }
+
+    private class MockTodoListPresenter implements TodoListPresenter {
```
ckatsube, line 79: Make sure to resolve warnings like making this class static, or not having unused exception declarations (in the two methods above)

src/main/java/services/event_creation/EventInfoFromReader.java
```diff
@@ -49,4 +50,17 @@ public LocalTime getEndTime() {
     public boolean getCompleted() {
         return eventReader.getCompleted();
     }
+
+    @Override
+    public boolean equals(Object o) {
```
ckatsube, line 55: why are adding methods purely for the test suite? or was there some larger intention when including this.
Also if you're going to override equals, might as well make it property-based, allowing for equality across other EventInfo instead of a wrapper based comparison with another concrete EventInfoFromReader. The current method exposes the underlying implementation underneath the abstraction, making it so that client code has to be aware that the two items it's comparing are EventInfoFromReader in order to compare equality (it should be something more abstract like the properties are the same)

src/main/java/services/task_presentation/TodoListInfoFromTaskReaders.java
```diff
@@ -31,4 +31,8 @@ public TodoListInfoFromTaskReaders(Map<Long, List<TaskReader>> taskReaders) {
             taskInfos.add(new TaskInfoFromTaskReader(tr));
         return taskInfos;
     }
+
+    public List<TaskReader> getTaskReaders() {
```
ckatsube, line 35: If you want to expose the Readers to the adapter and outward layers, might as well deprecate the -Info classes, but make that kind of change another PR. I think I was going to eventually go about doing this with the observable repositories PR but I hadn't committed to the idea yet

src/test/java/EventGetterTest.java
jenci2114: fixed

src/test/java/TaskOutputterTest.java
jenci2114: OK

ckatsube (approved):

src/main/java/services/event_creation/EventInfoFromReader.java
jenci2114: OK I deleted the equals method

src/main/java/services/task_presentation/TodoListInfoFromTaskReaders.java
jenci2114: Ok I removed the getter

PR #122, opened 2021-12-04
Email notification

JunruL:
[JavaMail API](https://javaee.github.io/javamail/#Download_JavaMail_Release)
Main Method is added in EmailNotificationPresenter for functionality testing, and will be deleted after the whole notification system is implemented.
For those whole review this PR, change the userEmail (lin507892@gmail.com) in `line 78` in `EmailNotificationPresenter` to your email, and run the main method.


src/main/java/services/notification_sending/EmailNotificationPresenter.java
```diff
@@ -0,0 +1,81 @@
...
+        this.enabled = enabled;
+        this.senderEmail = senderEmail;
+        this.password = password;
+        this.userEmail = userEmail;
+        this.subject = subject;
+    }
+
+    @Override
+    public void presentNotification(String message) {
+        //Get properties object
```
ckatsube, line 28: This kind of comment is more distracting than it is helpful. Comments should be things that are difficult to express in the code (e.g. why are you using "465" as the port?). If you really feel that a bundle of lines should be categorized, it's better self-documented as a private method. In this scenario, it's also a little more reusable that way because you can then make it into a factory method, taking in stuff like the port or SSL factory or host.

src/main/java/services/notification_sending/EmailNotificationPresenter.java
```diff
@@ -0,0 +1,81 @@
...
+        return this.userEmail;
+    }
+
+    public void setUserEmail(String userEmail) {
+        this.userEmail = userEmail;
+    }
+
+    public static void main(String[] args) {
+        EmailNotificationPresenter emailSender = new EmailNotificationPresenter(
+                true, "TimeTeam207@gmail.com",
```
ckatsube, line 77: Let's not have private info hard coded in like this. We can just use environment variables

src/main/java/services/notification_sending/EmailNotificationPresenter.java
```diff
@@ -0,0 +1,81 @@
+package services.notification_sending;
+
+import java.util.Properties;
+import javax.mail.*;
+import javax.mail.internet.*;
```
ckatsube, line 5: Change wildcard imports to explicit. You can update your IDE settings and make the auto bundling count like 999

src/main/java/services/notification_sending/EmailNotificationPresenter.java
```diff
@@ -0,0 +1,81 @@
...
+        //Get properties object
+        Properties props = new Properties();
+        props.put("mail.smtp.host", "smtp.gmail.com");
+        props.put("mail.smtp.socketFactory.port", "465");
+        props.put("mail.smtp.socketFactory.class",
+                "javax.net.ssl.SSLSocketFactory");
+        props.put("mail.smtp.auth", "true");
+        props.put("mail.smtp.port", "465");
+
+        //get Session
```
ckatsube, line 37: Fairly redundant comment when the next line says "Session get Instance"

Might be clearer if you describe why we need the anonymous class here (the `new javax.mail.Authenticator() {}`) and what said anonymous class is doing.

valyippee (approved):

src/main/java/services/notification_sending/EmailNotificationPresenter.java
```diff
@@ -0,0 +1,81 @@
...
+ * Keeps a record of whether the user enabled the email notification settings.
+ */
+public class EmailNotificationPresenter implements NotificationPresenter, SettingsRegistry {
+    private boolean enabled;
+    private String senderEmail;
+    private String password;
+    private String userEmail;
+    private String subject;
+
+    EmailNotificationPresenter(boolean enabled, String senderEmail, String password, String userEmail, String subject) {
```
ckatsube, line 18: this constructor is protected, not public.

src/main/java/services/notification_sending/EmailNotificationPresenter.java
```diff
@@ -0,0 +1,81 @@
...
+        //compose message
+        try {
+            MimeMessage msg = new MimeMessage(session);
+            msg.addRecipient(Message.RecipientType.TO,new InternetAddress(userEmail));
+            msg.setSubject(subject);
+            msg.setText(message);
+            //send message
+            Transport.send(msg);
+            System.out.println("message sent successfully");
+        } catch (MessagingException e) {throw new RuntimeException(e);}
```
ckatsube, line 54: Which part of the composing or sending actually throws this error? This try block is encompassing a number of 3rd party code so it's difficult to tell what you're actually expecting to be thrown.

Also I'm thinking that we either want to silence this exception or throw one of our own because otherwise this'll stop every other notification system from working. If we make our own exception then whatever class sends these notifs can just catch the error with just one exception (OCP), but I feel like in the end it'll just silence the exception there anyway.

src/main/java/services/notification_sending/EmailNotificationPresenter.java
```diff
@@ -0,0 +1,81 @@
...
+
+        //get Session
+        Session session = Session.getDefaultInstance(props,
+                new javax.mail.Authenticator() {
+                    protected PasswordAuthentication getPasswordAuthentication() {
+                        return new PasswordAuthentication(senderEmail, password);
+                    }
+                });
+
+        //compose message
```
ckatsube, line 45: I'm being a little harsh on comments today (sorry!) This comment isn't quite true since the following block both composes the message and also sends it. It doesn't contribute much to readability either as the following lines are pretty straightforward. In a sense it's actually a little confusing since the end of the method has been documented as 'composing the message' instead of 'sending the message.' I don't think I'd really want a `// compose and send message` comment either since as just mentioned this block is fairly straight forward already.

src/main/java/services/notification_sending/EmailNotificationPresenter.java
```diff
@@ -0,0 +1,81 @@
+package services.notification_sending;
+
+import java.util.Properties;
+import javax.mail.*;
+import javax.mail.internet.*;
+
+/**
+ * Sends out email notifications.
+ * Keeps a record of whether the user enabled the email notification settings.
```
ckatsube, line 9: will every notif presenter be responsible for checking if its enabled or not? Would this not be a violation of DRY

src/main/java/services/notification_sending/EmailNotificationPresenter.java
JunruL: This method will be removed later

src/main/java/services/notification_sending/EmailNotificationPresenter.java
JunruL: Yes, in the presentNotification method, enabled being true should lead to an early return.
Another way is to add and remove presenters in `observers` list of `NotificationTracker`.
We think record `enabled` in presenter itself might be easier to implement, since we don't need to create a new `presenter` object and add it to `observers` when enable it every time.

src/main/java/services/notification_sending/EmailNotificationPresenter.java
JunruL: Sorry I'm a bit lazy so I will just show the screenshoots. Based on these documentations, I don't think we are able to check and deal with some exceptions by ourselves.
截屏2021-12-04 上午10 23 05
截屏2021-12-04 上午10 23 16
截屏2021-12-04 上午10 23 22

emilychangg (approved):

jenci2114 (approved):

PR #124, opened 2021-12-04
added proxy for creating task and event

JunruL:
After introducing notification system, a new notification will be created when a new task/event is added.
We will use proxy design pattern for this extra work.
See the left bottom part of this [UML](https://app.diagrams.net/#G1d6-EMS59UJDCOAwQ8ZjQiKCUdbTaeNX0)
The `completeTask` method will be removed after #115 is merged.
[Action Failure](https://github.com/gradle/wrapper-validation-action/issues/40)

### DIFF
--- a/src/main/java/services/notification/NotificationAdder.java
+++ b/src/main/java/services/notification/NotificationAdder.java
@@ -7,7 +7,8 @@ public class NotificationAdder {
 
     /***
      * Create an event notification using the date from eventData and the event id eventId
-     * @param eventData the event data from the user
+     * @param eventData the event data passed down from the user interface
+     *                  when the user creates a new event
      * @param eventId the id of the event
      */
     public void createNotification(CalendarEventModel eventData, long eventId) {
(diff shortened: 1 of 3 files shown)


src/main/java/services/notification/NotificationAdder.java
```diff
@@ -0,0 +1,23 @@
+package services.notification;
+
+import services.event_creation.CalendarEventModel;
+import services.task_creation.TodoListTaskCreationModel;
+
+public class NotificationAdder {
+
+    /***
+     * Create an event notification using the date from eventData and the event id eventId
+     * @param eventData the event data from the user
```
valyippee, line 10: Instead of saying "event data from the user", how about "event data passed down from the user interface when the user creates a new event"?

I think it's clearer like that.

src/main/java/services/notification/NotificationAdder.java
```diff
@@ -0,0 +1,23 @@
...
+     * @param eventId the id of the event
+     */
+    public void createNotification(CalendarEventModel eventData, long eventId) {
+    }
+
+    /***
+     * Create a task notification using the date from taskData and the task id tasktId
+     * @param taskData the task data from the user
```
valyippee, line 18: same as the other comment about "data from the user"

valyippee (approved):

PR #125, opened 2021-12-04
Created view model factory 

valyippee:
1. Created view model factory to return view models, to be injected into fx-controllers (the model class that view models depend on are not included, but a sample/ dummy model is included for me to try out/ test)
2. Added a `ViewModelBindingController` interface for all controllers who need a view model to implement (actually only the Navigation controllers do not need implement this)
3. Created `AdaptedFXMLLoader` to not only load the fxml content and instantiate the associated controller, but also **_inject the view model into the controller_** after instantiation (by calling the `init()` method in `ViewModelBindingController` interface. this is also why i created the interface). A problem was that `NavigationHelper` is static, so I couldnt inject the `ViewModelFactory` into it for it to get the various view models to be injected into the fx-controllers. So I let `NavigationHelper` hold an instance of `AdaptedFXMLLoader` and let `AdaptedFXMLLoader` create the `ViewModelFactory` which it uses to inject view models into fx-controllers whenever being called. But I'm not sure if having `ViewModelFactory` being constructed in `AdaptedFXMLLoader` is a good idea (versus instantiating it in `GUIDriver`, for example).

### DIFF
--- a/src/main/java/gui/view_model/MonthlyCalendarViewModel.java
+++ b/src/main/java/gui/view_model/MonthlyCalendarViewModel.java
@@ -1,0 +1,48 @@
+package gui.view_model;
+
+import com.calendarfx.model.Entry;
+import data_gateway.event.EventReader;
+import gui.model.EventModelManager;
+import javafx.collections.FXCollections;
+import javafx.collections.ObservableList;
+
+import java.time.LocalDate;
+import java.util.ArrayList;
+import java.util.List;
+
+public class MonthlyCalendarViewModel extends ViewModel {
+
+    private final EventModelManager manager;
+
+    private final ObservableList<Entry<String>> entryList;
+
+    public MonthlyCalendarViewModel(EventModelManager manager) {
+        this.manager = manager;
+
+        List<Entry<String>> newList = new ArrayList<>();
+
+        List<EventReader> eventReaderList = this.manager.getEvents();
+        for (EventReader er : eventReaderList) {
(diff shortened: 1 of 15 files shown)


src/main/java/gui/view_model/MonthlyCalendarViewModel.java
```diff
@@ -0,0 +1,48 @@
...
+
+    }
+
+    public ObservableList<Entry<String>> getEntryList() {
+        System.out.println(this.entryList);
+        return this.entryList;
+    }
+
+    private Entry<String> eventReaderToEntry(EventReader eventReader) {
+        Entry<String> entry = new Entry<>(eventReader.getName());
```
TheXanMan321, line 40: I'm getting a warning here about a duplicated string, not sure what the issue is

src/main/java/gui/view_model/MonthlyCalendarViewModel.java
```diff
@@ -0,0 +1,48 @@
...
+        entry.changeStartTime(eventReader.getStartTime());
+        entry.changeEndTime(eventReader.getEndTime());
+        LocalDate date = LocalDate.of(2021, 12, 8);
```
TheXanMan321, line 43: Could you change this so that it uses the events date?

TheXanMan321 (commented):
This is really good! The only thing I think is missing is that events all get loaded on the same day.

PR #126, opened 2021-12-05
GUI refactoring

valyippee:
1. Created view model factory to return view models, to be injected into fx-controllers (the model class that view models depend on are not included, but a sample/ dummy model is included for me to try out/ test)
2. Added a ViewModelBindingController interface for all controllers who need a view model to implement (actually only the Navigation controllers do not need to implement this)
3. Added `InstanceMapper` which maps a controller class to its corresponding view model class. `NavigationHelper` holds an instance of this mapper and uses it to determine which view models to inject into each controller upon instantiation.

### DIFF
--- a/src/main/java/gui/GUIDriver.java
+++ b/src/main/java/gui/GUIDriver.java
@@ -1,17 +1,29 @@
 package gui;
 
+import data_gateway.event.CalendarManager;
+import data_gateway.event.EventEntityManager;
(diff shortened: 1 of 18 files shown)


ckatsube (approved):

TheXanMan321 (approved):
I'm a little concerned the ViewModelBindingController might violate interface segregation, but I don't think it's an urgent issue

PR #127, opened 2021-12-05
Full pomodoro GUI

TheXanMan321:


ckatsube:
Bumping my old comment of the time fields should have little validation flag and disabled start button instead of a full blown error message popup (look up "jetbrains ui validation script" for an example of what I'm talking about. The example is from a different UI framework). Also it'd be nice to have the text fields actually appear uninteractable while the timer is running, not just be unclickable. It's a bit misleading to have a field that looks interactable but not be clickable.

It would also be nice to have a little visual indication of what the current session is/how many sessions are left. Something to tell the user where in the Pomodoro technique they are.

Other than that nice work!

jenci2114 (approved):

PR #128, opened 2021-12-05
Added save feature for notifications

TheXanMan321:
Small change, adding save method for Notification JSON adapter

### DIFF
--- a/src/main/java/data_gateway/notification/JsonNotificationAdapter.java
+++ b/src/main/java/data_gateway/notification/JsonNotificationAdapter.java
@@ -6,6 +6,8 @@ import com.google.gson.stream.JsonWriter;
 import entity.Notification;
 
 import java.io.IOException;
+import java.time.Duration;
+import java.time.LocalDateTime;
 
 // Based on GSON type adapter documentation from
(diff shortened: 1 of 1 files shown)


valyippee (approved):

src/main/java/data_gateway/notification/JsonNotificationAdapter.java
```diff
@@ -24,7 +26,30 @@ public void write(JsonWriter jsonWriter, Notification notification) throws IOExc
 
 
     @Override
-    public Notification read(JsonReader jsonReader) {
+    public Notification read(JsonReader jsonReader) throws IOException {
+        long id = 0;
```
ckatsube, line 30: I'm sure that we'll never actually get a 0 id with the Snowflake, but because 0 is actually a valid id it would be nice to have either something invalid (maybe negative?) or null. That way this seems less like a  "default value" and more of a "special invalid value."
It also helps in handling invalid cases if that ever needs to be a thing (but for the most part it's implicit that the initial values will be replaced).

ckatsube (approved):

jenci2114 (approved):

PR #133, opened 2021-12-06
Gui calendar propagating data for event creation

valyippee:
New events created on the calendar page (by double clicking) will be propagated to the repositories, and a save option is provided to persist the data.

Currently does not support updating of events (so after event creation, any changes of event in the gui is not propagated). Will add it in another pr.


TheXanMan321 (approved):

src/main/java/gui/view_model/MonthlyCalendarViewModel.java
```diff
@@ -1,38 +1,120 @@
...
+        this.entryList.addListener((ListChangeListener<Entry<String>>) c -> {
+            while (c.next()) {
+                if (c.wasAdded()) {
+                    Entry<String> entry = c.getAddedSubList().get(0);
+
+                    // if entryList and entryToEventIdMapping has the same size,
+                    // entry added to entryList is due to changes in data in the
+                    // repository and can be ignored as the mapping has been added.
+
+                    if (entryList.size() == entryToEventIdMapping.size() + 1) {
```
TheXanMan321, line 42: I think this only gets updated when an event is created, not when it's attributes are changed, is there a way to fix this?

PR #135, opened 2021-12-06
Finished Task addition from GUI

jenci2114:


ckatsube:
A lot of the basic stuff is down, nice work!

Make sure to constraint the inputs of the field.
For the time fields, only numbers (and up to 2) should be included. Also try including default values (00:00 might make sense)
For the duration, similarly only allow numbers.

Although the fields should only allow to be populated by valid inputs, you should still have an error window when unable to create the task. This isn't a feature for the user, but more of a failsafe measure in case your field validation wasn't sufficient.

Would be nice to have the due date box border change colors based on if it contains a valid due date or not. Or make the field unchangeable except when using the date picker.

Subtasks button doesn't seem to work for me (I click 'add' and there's no visible change).

ckatsube:
Have you looked into the TextFormatter?
https://docs.oracle.com/javase/8/javafx/api/javafx/scene/control/TextFormatter.html
If it feels too unwieldy then that's OK too.

Last suggestion from me:
since the subtasks are laid out separately from the rest of the form elements, it would be nice to have the add/remove buttons grouped with the subtask list. Since there's a bounding box that appears when you delete all the subtasks (might be a bug but) I thought of the idea of just permanently showing that bounding box and having the add/remove buttons underneath it. If you add more subtasks than the box supports height, then just have the box be vertically scrollable. Of course feel free to implement your own layout choice if you have an idea that appeals to you more.

jenci2114:
Now the bounding box is permanently shown and the add/remove buttons are on top of the box

JunruL:
Creating tasks without deadline should be allowed. (currently task fails to be created without due date)
Not sure whether duration should be filled for creating a task. (currently task fails to be created without duration)
It seems that the tasks created could not be saved to database.
Note that there should be some update on this pr after #138 is merged (new task should be inserted into the todo list based on the due date)

src/main/resources/addTaskPage.fxml
```diff
@@ -0,0 +1,63 @@
...
+<?import javafx.scene.text.Font?>
+<?import javafx.scene.text.Text?>
+
+<AnchorPane maxHeight="-Infinity" maxWidth="-Infinity" minHeight="-Infinity" minWidth="-Infinity" prefHeight="800.0" prefWidth="1000.0" xmlns="http://javafx.com/javafx/17" xmlns:fx="http://javafx.com/fxml/1" fx:controller="gui.view.AddTaskPageController">
+    <children>
+        <AnchorPane fx:id="mainBackground" layoutX="294.0" prefHeight="800.0" prefWidth="740.0" style="-fx-background-color: #E5E5E5;" AnchorPane.bottomAnchor="0.0" AnchorPane.leftAnchor="50.0" AnchorPane.rightAnchor="0.0" AnchorPane.topAnchor="0.0">
+            <children>
+            <AnchorPane layoutX="66.0" layoutY="44.0" prefHeight="661.0" prefWidth="809.0" style="-fx-background-color: #FFFFFF;">
+               <children>
+                  <DialogPane fx:id="message" layoutX="45.0" layoutY="448.0" />
```
JunruL, line 19: We can make this box transparent when there is no message.
I was confused why there is such a box before I failed to created a task and the message shows up.

src/main/resources/taskBar.fxml
```diff
@@ -0,0 +1,64 @@
...
+                  <Font size="20.0" />
+               </font>
+            </Text>
+            <TextField fx:id="dueTimeMinutes" layoutX="340.0" layoutY="228.0" prefHeight="26.0" prefWidth="39.0" />
+            <TextField fx:id="duration" layoutX="281.0" layoutY="305.0" prefHeight="26.0" prefWidth="39.0" />
+            <ToggleButton fx:id="completed" layoutX="281.0" layoutY="388.0" mnemonicParsing="false" text="Yes" />
+            <Button fx:id="addSubtask" layoutX="282.0" layoutY="490.0" mnemonicParsing="false" text="Add" />
+            <Button fx:id="scheduleAsEvent" layoutX="111.0" layoutY="609.0" mnemonicParsing="false" text="Schedule as Event" />
+            <Button fx:id="updateTask" layoutX="286.0" layoutY="609.0" mnemonicParsing="false" text="Update" />
+            <Button fx:id="deleteTask" layoutX="115.0" layoutY="688.0" mnemonicParsing="false" text="Delete Task" />
```
JunruL, line 54: We do not have task remover on the back end right?

jenci2114:
the deadline input is now optional

src/main/resources/addTaskPage.fxml
jenci2114: ok fixed

src/main/resources/taskBar.fxml
ckatsube: We don't. We can add one though, it shouldn't be too hard. Do you think we should?

jenci2114: The task modification part is not yet implemented. If we don't have one, this feature will be removed when the task modification part (taskBar) is completed

ckatsube: task updating has been added in the backend (did you mean in the GUI?). Task removing is the one that hasn't been added yet

jenci2114: Yeah I meant the GUI

jenci2114: I'll do that for the next PR

JunruL: I think it would be good if we have deleted choice.

JunruL (approved):

PR #136, opened 2021-12-06
GUI Main Page

JunruL:


ckatsube:
I like the lists and the overall layout of the main page.

Make sure to use some more responsive layouts. Currently resizing the screen causes the components to go haywire.
Also I noticed that the list items were selectable, is there a reason for that? If so and it just hasn't been implemented yet, no problem! Just wanted to bring to attention that when it's selectable, the user kind of expects there to be a reason.

I know that properly filtering the events/tasks to display the relevant ones is a feature currently being worked on. So other than that, nothing much to add. Great work 👍

JunruL:
@ckatsube 
It seems that currently all sub pages have the resizing screen issue... I will can look into it but now I have some other tasks to do. 
AS for the selectable items, I think it's a feature of `JFXListView`. Those listed tasks and event should bee only for user to view.

ckatsube:
Got it. I think it's a good idea to look into the resizing issue as soon as you have all the pages working. If you don't have time, then it might be a good idea to consider disabling the window resizing grip.

PR #138, opened 2021-12-07
GUI - add task id and sort todo list

JunruL:
- In the todo list page, each task component should have associated id. This PR added the id.
- This PR also sort the todo list by deadline. Those tasks without deadlines will be put after the ones with deadlines.


src/main/java/gui/viewmodel/TodoListPageViewModel.java
```diff
@@ -3,30 +3,74 @@
...
     }
 
-    private ObservableMap<String, String> generateTaskNameDeadlineMap(List<TaskReader> taskList) {
-        ObservableMap<String, String> nameDeadlineMap = FXCollections.observableHashMap();
-        for (TaskReader taskReader : taskList) {
+    /***
+     * helper function to sort the taskReaderList based on the deadline
+     * @param taskReader to inserted to taskReadList
+     */
+    private void insertTaskReader(TaskReader taskReader) {
```
jenci2114, line 45: For the sorting part, we could use java's default sorting with lambda expression

src/main/java/gui/viewmodel/TodoListPageViewModel.java
```diff
@@ -3,30 +3,74 @@
...
 import datagateway.task.TaskReader;
 import javafx.collections.FXCollections;
-import javafx.collections.ObservableMap;
+import javafx.collections.ObservableList;
 
+import java.time.LocalDateTime;
 import java.time.format.DateTimeFormatter;
 import java.time.format.FormatStyle;
-import java.util.List;
+import java.util.*;
```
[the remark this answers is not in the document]
JunruL: Yes, I wil change it

src/main/java/gui/viewmodel/TodoListPageViewModel.java
JunruL: Right, but since there might be non-deadline tasks, I did it manually since I'm not sure whether lambda can deal with this.

jenci2114: ok. If lambda does not work we can leave it like this

jenci2114 (approved):

emilychangg (approved):

ckatsube (approved):

PR #140, opened 2021-12-07
pomodoro persistence between scenes

emilychangg:
- Allows the user to leave the pomodoro page and come back without the timer being restarted
- Removed TimeFormatter class as per Craig's comment on previous PR
- Fixed bug where timer couldn't handle inputs greater than 60 minutes
- Fixed bug where resetting the clock didn't actually reset the clock


src/main/java/entity/ConsolePomodoroTimer.java
```diff
@@ -0,0 +1,35 @@
+package entity;
+
+public class ConsolePomodoroTimer {
```
valyippee, line 3: Our entity should not depend on whether we are using the console or GUI. Why is there a need for a new class?

valyippee (approved):

src/main/java/entity/ConsolePomodoroTimer.java
emilychangg: oh I was just renaming some classes so that I could use their old names for classes in the GUI. This used to be PomodoroTimer but I changed a lot about the pomodoro to make it work for the GUI so the old PomodoroTimer is no longer relavent. I didn't want to fully delete the class though because I remember our TA saying that we should still hand in our console application and these old pomodoro classes are necessary for the console app to work.

PR #143, opened 2021-12-07
Made Events Better Linked To Its Task

ckatsube:
Yepp this is a PR with a bunch of different changes. A lot of things going on here that I don't think I had time to make into smaller PRs

1) extracted Event's explicit reference to task to an id instead
2) made the event manager use the task manager to create a task for new events (this is new behavior. I can remove this specific part)
2.5) let the event manager use an existing task when creating creating an event from a task
3) added new request class for said event-from-task creating
4) removed Event's direct name and completion status. Now lobbies off of its associated task's name and completion status.
5) updated `HashSet` variable types along Event/Task creation to `Set`
6) removed task manager's adding via dto to using more specific data.
7) updated JSON saving so it works the same as before, except Events now store taskId instead of an entire task.

Changes tested:
- adding a new event will show both the event and the associated task in the same format as before.
- creating a task from an event and then editing the completion status of the task or event causes both to change.
- loading and saving events and tasks


src/main/java/datagateway/event/EventToEventReader.java
```diff
@@ -8,9 +8,13 @@
...
 public class EventToEventReader implements EventReader{
     private final Event event;
+    private final String name;
+    private final boolean completed;
 
-    public EventToEventReader(Event event){
+    public EventToEventReader(Event event, String name, boolean completed){
         this.event = event;
+        this.name = name;
+        this.completed = completed;
```
jenci2114, line 17: why shouldn't we obtain the event name and completion from the event itself?

src/main/java/datagateway/task/TodoListManager.java
```diff
@@ -10,7 +8,7 @@
 
 public interface TodoListManager {
 
-    long addTask(TodoListTaskCreationModel taskData);
+    long addTask(String name, Duration duration, LocalDateTime deadline, List<String> subtasks);
```
jenci2114, line 11: may need to change how the front end VM saves task

src/main/java/datagateway/event/EventToEventReader.java
ckatsube: Event doesn't have this info anymore since it doesn't explicitly store a task. It stores the taskId which is used to get the name and completed status.

src/main/java/entity/Calendar.java
```diff
@@ -14,7 +14,7 @@ public Calendar(String name, List<Event> events) {
         TodoList todoList = new TodoList();
 
         for(Event event : events) {
-            Task task = new Task(event.getId(), event.getEventName());
+            Task task = new Task(event.getId(), "");
```
valyippee, line 17: why are we making the task name an empty string? we can give it a default name as being the same as the associated event.

jenci2114 (approved):

src/main/java/datagateway/event/EventToEventReader.java
valyippee: it would make sense for event to have its own name, separated from task. e.g. you might have "complete tutorial worksheet" as your task, and "meeting with study group to discuss tutorial problems" as your event. but i guess its a small thing and we dont need to make the change.

src/main/java/consoleapp/eventadapters/EventController.java
```diff
@@ -16,6 +17,7 @@
 import java.util.HashMap;
 import java.util.HashSet;
```
valyippee, line 18: extra import

valyippee (approved):

src/main/java/entity/Calendar.java
ckatsube: ah that's because the event entity itself doesn't hold its name anymore, and the calendar doesn't even do anything now so I just arbitrarily changed the value.

PR #145, opened 2021-12-07
Removed redundant classes in GUI

valyippee:
1. removed basic page and controller since we dont need a reference page anymore
2. removed `dailyCalender`'s fxml and controller since we decided not to implement that anymore
3. moved `ClockController` to `view` since it should belong there


src/main/java/gui/GUIDriver.java
```diff
@@ -24,7 +24,7 @@
     @Override
     public void start(Stage primaryStage) throws Exception{
         configure();
-        Parent root = FXMLLoader.load(Objects.requireNonNull(getClass().getResource("/basicPage.fxml")));
+        Parent root = FXMLLoader.load(Objects.requireNonNull(getClass().getResource("/monthlyCalendar.fxml")));
```
TheXanMan321, line 27: Are we going to change `monthlyCalendar.fxml` to the main page later?

TheXanMan321 (approved):

ckatsube (approved):

src/main/java/gui/GUIDriver.java
valyippee: No it will be the `mainPage.fxml` in junru's pr, but since there isnt a main right now, I used the monthly one first.

PR #147, opened 2021-12-07
Added/ changed methods in request boundaries

valyippee:
Previously, the request boundaries for event getting and task getting was not standardized e.g. in event getter, we are getting `List<HashMap>` instead of `List<EventInfo>` when getting all events, and in task getter, we didnt have the method to get all tasks.

### DIFF
--- a/src/main/java/consoleapp/eventadapters/EventController.java
+++ b/src/main/java/consoleapp/eventadapters/EventController.java
@@ -13,6 +13,7 @@ import services.strategybuilding.DatesForm;
 import java.io.IOException;
 import java.time.Duration;
 import java.time.LocalTime;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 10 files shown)


src/main/java/consoleapp/eventadapters/EventController.java
```diff
@@ -84,6 +85,17 @@ public EventInfo getEventByName(String name) {
     }
 
     public List<HashMap<String, String>> getEvents() {
-        return eventGetter.getEvents();
+        List<HashMap<String, String>> hashMapList = new ArrayList<>();
+        List<EventInfo> eventInfos = eventGetter.getEvents();
+        for (EventInfo info : eventInfos) {
+            HashMap<String, String> infoMap = new HashMap<>();
```
JunruL, line 91: The id of event might be needed?

src/main/java/services/taskpresentation/TaskGetter.java
```diff
@@ -31,4 +32,20 @@ public TaskInfo getTaskById(Long id) {
         return null;
     }
 
+    /**
+     * Get all tasks.
+     * @return a list of TaskInfo
+     */
+    @Override
+    public List<TaskInfo> getTasks() {
```
JunruL, line 40: Seems there is no corresponding update in Task controller

src/main/java/consoleapp/eventadapters/EventController.java
valyippee: The previous implementation had the `EventController` return a hashmap without the ids. I didn't want to change that coz this pr was just about changing the interfaces of request boundaries.

src/main/java/services/taskpresentation/TaskGetter.java
valyippee: Yup because `TaskController` did not use this method

ckatsube (approved):

src/main/java/consoleapp/eventadapters/EventController.java
JunruL: Right, I thought it has something to do with gui

src/main/java/services/taskpresentation/TaskGetter.java
JunruL: OK so this pr is just about refactoring

JunruL (approved):

PR #149, opened 2021-12-07
Separated dependencies in GUI

valyippee:
Currently, the view models use the repositories directly. However, they should be interacting with the use cases who will then do validation checks, processing etc, and propagate data to the repos. So I've separated out these dependencies and made the view model unaware of the repos.

I've only changed the calendar view models. Will leave the changes for todolist view model to @JunruL.


src/main/java/gui/utility/EventHelper.java
```diff
@@ -14,20 +15,20 @@
...
 
-    public static Entry<String> eventReaderToEntry(EventReader eventReader) {
-        Entry<String> entry = new Entry<>(eventReader.getName());
-        entry.changeStartTime(eventReader.getStartTime());
-        entry.changeEndTime(eventReader.getEndTime());
+    public static Entry<String> eventInfoToEntry(EventInfo eventInfo) {
+        Entry<String> entry = new Entry<>(eventInfo.getName());
+        entry.changeStartTime(eventInfo.getStartTime());
+        entry.changeEndTime(eventInfo.getEndTime());
         LocalDate date = LocalDate.of(2021, 12, 8);
```
ckatsube, line 22: This might be a forgotten constant. I'm guessing it should be injected into this method instead?

src/main/java/gui/viewmodel/MonthlyCalendarViewModel.java
```diff
@@ -1,38 +1,48 @@
...
-        this.entryList = FXCollections.observableArrayList(EventHelper.eventReaderToEntry(eventReaderList));
+    public MonthlyCalendarViewModel(CalendarEventCreationBoundary eventAdder,
+                                    CalendarEventRequestBoundary eventGetter,
+                                    UpdateEventBoundary eventUpdater,
+                                    EventSaver eventSaver) {
+        this.eventAdder = eventAdder;
+        this.eventGetter = eventGetter;
+        this.eventUpdater = eventUpdater;
+        this.eventSaver = eventSaver;
+        List<EventInfo> eventInfoList = new ArrayList<>(); // TODO: change this to be getting events from EventGetter
```
ckatsube, line 35: How come you can't do this todo already? Is there some change that we need to do first

src/main/java/gui/utility/EventHelper.java
valyippee: right it was for me to test the view. I will include the change in another pr with the calendar updates and creation.

src/main/java/gui/viewmodel/MonthlyCalendarViewModel.java
valyippee: got it done now! was waiting for another pr to be merged

ckatsube (approved):

jenci2114 (approved):

src/main/java/gui/viewmodel/TodoListPageViewModel.java
```diff
@@ -12,87 +13,78 @@
...
                 i = i + 1;
             }
         }
-        taskReaderList.add(taskReader);
+        taskInfoList.add(taskInfo);
     }
 
     /***
-     * Update TaskInfoList to make it consistent with taskReaderList
+     * Update viewInfoList to make it consistent with taskReaderList
```
JunruL, line 62: to make it consistent with taskIntoList

src/main/java/gui/viewmodel/TodoListPageViewModel.java
```diff
@@ -12,87 +13,78 @@
...
+            this.viewInfoList.add(taskViewInfo);
         }
     }
 
     public ObservableList<Map<String, String>> getTaskInfoList() {
-        return taskInfoList;
+        return viewInfoList;
     }
 
     public void handleCreation(TaskReader taskReader) {
```
JunruL, line 90: Still using `taskReader` here

src/main/java/gui/viewmodel/MonthlyCalendarViewModel.java
```diff
@@ -3,26 +3,37 @@
 import com.calendarfx.model.Entry;
 import datagateway.Observer;
 import datagateway.event.EventReader;
```
JunruL, line 5: Using `EvntInfo`  instead?

src/main/java/gui/viewmodel/ViewModelFactory.java
```diff
@@ -4,37 +4,58 @@
...
-    public ViewModelFactory(ObservableRepositoryFactory repositoryFactory) {
-        this(repositoryFactory.makeEventRepository(), repositoryFactory.makeTaskRepository());
+    public ViewModelFactory(ObservableRepositoryFactory repositoryFactory, ServicesFactory servicesFactory) {
+        this.servicesFactory = servicesFactory;
+        eventRepository = repositoryFactory.makeEventRepository();
+        taskRepository = repositoryFactory.makeTaskRepository();
     }
 
     public MonthlyCalendarViewModel getMonthlyCalendarViewModel() {
+        if (monthlyCalendarViewModel == null) {
```
JunruL, line 27: Why having this condition check?

valyippee: becoz if it is null, it means that it has not been initialized. if it is not null, we will return the one that has been initialized. this ensures that we are always getting the same view model

src/main/java/gui/viewmodel/TodoListPageViewModel.java
valyippee: this task reader is received when the observable repo updates its observers. wasnt part of the refactoring.

src/main/java/gui/viewmodel/MonthlyCalendarViewModel.java
valyippee: both will be needed as we are receiving an EventReader from the observable repo when getting updates.

src/main/java/gui/viewmodel/TodoListPageViewModel.java
valyippee: right good catch. i will update it in another pr.

PR #153, opened 2021-12-07
added deletion use case

ckatsube:
included deletion observing and example in ApplicationDriver


src/main/java/services/servicesfactory/NotificationServiceFactory.java
```diff
@@ -70,6 +72,12 @@ public UpdateEventBoundary makeEventUpdater() {
         return innerFactory.makeEventUpdater();
     }
 
+    @Override
+    public EventDeletionBoundary makeEventDeleter() {
+        //TODO: develop notification-deletion family event deleter
```
valyippee, line 77: Can you add an issue for this? I probably can do it in the notification pr

valyippee (approved):

jenci2114 (approved):

PR #154, opened 2021-12-07
Main page gui

halfxblood:
The two data files will be deleted before merging.
Having them in this PR makes it easier for people to review.


src/main/java/gui/view/MainPageController.java
```diff
@@ -0,0 +1,159 @@
...
+    /**
+     * Method listening to updates from the {@link MainPageViewModel}.
+     * @param readers the lists of updated readers
+     */
+    private void acceptReaders(MainPageViewModel.Readers readers) {
+        updateListView(taskListView, formatTaskReaders(readers.taskReaders));
+        updateEventListView(eventListView, formatEventReaders(readers.eventReaders));
+    }
+
+    private Map<String, String> formatTaskReaders(List<TaskReader> taskReaders) {
```
JunruL, line 115: View should not interact with data gateway

src/main/java/gui/view/MainPageController.java
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    private String formatDuration(LocalDateTime localDateTime) {
+        return localDateTime.format(
+                DateTimeFormatter.ofLocalizedDateTime(
+                        FormatStyle.MEDIUM,
+                        FormatStyle.SHORT));
+    }
+
+    private Map<String, List<String>> formatEventReaders(List<EventReader> eventReaders) {
```
JunruL, line 135: Should not interact with data gateway.

src/main/java/gui/view/MainPageController.java
```diff
@@ -0,0 +1,159 @@
...
+        }
+        for (TaskReader task: taskReaders) {
+            String taskName = task.getName();
+            String deadline = formatDuration(task.getDeadline());
+            taskMap.put(taskName, deadline);
+        }
+        return taskMap;
+    }
+
+    private String formatDuration(LocalDateTime localDateTime) {
```
JunruL, line 128: In my understanding, view is just receiving something to display? So probably move the converting LocalDateTime to String to view model?

src/main/java/gui/view/MainPageController.java
```diff
@@ -0,0 +1,159 @@
...
+        updateEventListView(eventListView, formatEventReaders(this.viewModel.getRelevantEvents()));
+
+    }
+
+    /**
+     * Note: removes visibility of all previously rendered labels in the list view before flushing it out.
+     * @param listView the list view to update
+     * @param map a mapping of name and time for today's relevant items
+     */
+    private void updateListView(JFXListView<HBox> listView, Map<String, String> map) {
```
JunruL, line 61: - method name `updateTaskListView`?

- Since this is for task list, I think the parameter `JFXListView<HBox> listView` is not necessary

src/main/java/gui/view/MainPageController.java
```diff
@@ -0,0 +1,159 @@
...
+        listView.refresh();
+
+    }
+
+    /**
+     * Note: removes visibility of all previously rendered labels in the list view before flushing it out.
+     * @param eventListView the list view to update
+     * @param map a mapping of name and time for today's relevant items
+     */
+    private void updateEventListView(JFXListView<HBox> eventListView, Map<String, List<String>> map) {
```
JunruL, line 83: `eventListView` not necessary, same as above

src/main/java/gui/view/MainPageController.java
```diff
@@ -0,0 +1,159 @@
...
+        eventListView.getItems().forEach(i -> i.setVisible(false));
+        eventListView.getItems().clear();
+        for (Map.Entry<String, List<String>> eventInfo : map.entrySet()) {
+            Label eventName = new Label(eventInfo.getKey());
+            Label startTime = new Label("Start Time is " + eventInfo.getValue().get(0));
```
JunruL, line 88: Rather than display "Start Time is...     End Time is ....", I think it can just display "`SartTime` - `EndTime`"

src/main/java/gui/viewmodel/MainPageViewModel.java
```diff
@@ -0,0 +1,163 @@
...
+import java.util.List;
+import java.util.ArrayList;
+import java.util.function.Consumer;
+
+
+public class MainPageViewModel extends ViewModel implements Runnable{
+    private final ObservableTaskRepository taskRepository;
+    private final ObservableEventRepository eventRepository;
+    private final List<TaskReader> relevantTasks = new ArrayList<>();
+    private final List<EventReader> relevantEvents = new ArrayList<>();
```
JunruL, line 19: Should be `-info` instead of `-reader`. `-reader` is gateway

src/main/resources/style.css
```diff
@@ -0,0 +1,18 @@
+.title-label{
```
JunruL, line 1: Make the style name more specific since this file is under the resources folder, which organizes the whole fxml files

src/main/resources/setting.fxml
```diff
@@ -0,0 +1,68 @@
...
+<?import javafx.scene.layout.AnchorPane?>
+<?import javafx.scene.layout.HBox?>
+<?import javafx.scene.text.Font?>
+
+
+<AnchorPane maxHeight="-Infinity" maxWidth="-Infinity" minHeight="-Infinity" minWidth="-Infinity" prefHeight="800.0" prefWidth="1000.0" xmlns="http://javafx.com/javafx/17" xmlns:fx="http://javafx.com/fxml/1" fx:controller="gui.view.SettingsController">
+   <children>
+      <AnchorPane fx:id="mainBackground" layoutX="42.0" prefHeight="800.0" prefWidth="950.0" style="-fx-background-color: #E5E5E5;" AnchorPane.leftAnchor="50.0">
+         <children>
+            <HBox layoutX="130.0" layoutY="148.0" prefHeight="76.0" prefWidth="319.0" AnchorPane.leftAnchor="130.0" AnchorPane.topAnchor="148.0">
```
JunruL, line 18: Though I thought having HBox is redundant, we can still keep it.
Also I notice that your main page and setting page can fit the screen when I maximize the window (though it move to the left top corner, it's still better than other pages we have).
So if you have time, it would be good to fix that bugs for other pages

src/main/java/gui/view/MainPageController.java
```diff
@@ -0,0 +1,159 @@
...
+import javafx.fxml.Initializable;
+import javafx.scene.control.Label;
+import javafx.scene.layout.HBox;
+import javafx.scene.text.Font;
+
+import java.net.URL;
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.FormatStyle;
+import java.util.*;
```
ckatsube, line 20: change wildcard import to explicit imports

src/main/java/gui/viewmodel/MainPageViewModel.java
```diff
@@ -0,0 +1,163 @@
...
+    private final List<Consumer<Readers>> observers = new ArrayList<>();
+
+    public MainPageViewModel(ObservableTaskRepository observableTaskRepository, ObservableEventRepository observableEventRepository){
+        this.taskRepository = observableTaskRepository;
+        this.eventRepository = observableEventRepository;
+
+        observableTaskRepository.addCreationObserver(this::handleCreation);
```
ckatsube, line 26: repository binding was implemented during construction in the #149 PR. No need to do this anymore. Instead get the `TaskRequestBoundary` or whatever the return type of `serviceFactory.makeTaskGetter()` is, and use that instead of the repository.

src/main/java/gui/utility/NavigationHelper.java
```diff
@@ -117,6 +119,9 @@ public static void enterPomodoroPage(MouseEvent event) throws IOException {
      * @throws IOException if the resource file cannot be found
      */
     public static void enterSettingsPage(MouseEvent event) throws IOException {
+        FXMLLoader loader = new FXMLLoader();
+        loader.setLocation(Objects.requireNonNull(NavigationHelper.class.getResource("/setting.fxml")));
```
valyippee, line 123: can you include files related to the settings page here too?

JunruL: I asked @halfxblood  to open a new PR for setting.
I think he is working on that branch

build.gradle
```diff
@@ -31,4 +31,6 @@ test {
     testLogging {
         events "passed", "skipped", "failed"
     }
-}
\ No newline at end of file
+}
+sourceCompatibility = JavaVersion.VERSION_14
+targetCompatibility = JavaVersion.VERSION_14
```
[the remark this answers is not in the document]
ckatsube: We decided to use version 16, didn't we? also I don't think we need to specify the compatability.

src/main/java/gui/view/MainPageController.java
```diff
@@ -0,0 +1,162 @@
...
+
+            String startTime = event.getStartTime().format(
+                    DateTimeFormatter.ofLocalizedTime(
+                            FormatStyle.SHORT));
+
+            String endTime = event.getEndTime().format(
+                    DateTimeFormatter.ofLocalizedTime(
+                            FormatStyle.SHORT));
+            list.add(startTime);
+            list.add(endTime);
```
ckatsube, line 156: How about just adding the `startTime + " - "  + endTime` here instead of up in line 100 when updating the view? That way the map can just be a `Map<String, String>` and you won't have to worry about doing the `get(0)` `get(1)`

src/main/java/gui/viewmodel/MainPageViewModel.java
```diff
@@ -0,0 +1,179 @@
...
+    private final List<TaskInfo> relevantTasks = new ArrayList<>();
+    private final List<EventInfo> relevantEvents = new ArrayList<>();
+    private final List<Consumer<Infos>> observers = new ArrayList<>();
+
+    public MainPageViewModel(TodoListRequestBoundary taskGetter, CalendarEventRequestBoundary eventGetter){
+        this.taskRepository = taskGetter;
+        this.eventRepository = eventGetter;
+
+        updateRelevantTasks();
+        updateRelevantEvents();
```
ckatsube, line 36: Seems like these two are always paired together. Might be able to combine or wrap these methods together, which would also allow you to call the observer notify after update both lists instead of once after updating each list.

ckatsube (approved):

src/main/java/gui/utility/NavigationHelper.java
TheXanMan321, line 123: Does this setting page exist yet?

valyippee: it is not in this pr, but i think tahseen is still making adjustments to it

build.gradle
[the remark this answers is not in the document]
jenci2114: I had this issue where IntelliJ told me to add those lines, but if you use Java 16 as your JDK (rather than coretto-11) you shouldn't need those lines

PR #162, opened 2021-12-08
Completed Tasks have strikethrough

JunruL:


valyippee:
How do we check a task as completed?

valyippee (approved):

PR #164, opened 2021-12-08
Settings gui

halfxblood:


src/main/java/gui/GUIDriver.java
```diff
@@ -3,10 +3,7 @@
 import com.sun.source.tree.BreakTree;
 import gui.utility.InstanceMapper;
 import gui.utility.NavigationHelper;
-import gui.view.AddTaskPageController;
-import gui.view.MonthlyCalendarController;
-import gui.view.TodoListPageController;
-import gui.view.WeeklyCalendarController;
+import gui.view.*;
```
jenci2114, line 6: Avoid wild card import

jenci2114: You could change your IDE setting so it doesn't automatically merge imports into a wild card import

src/main/java/gui/viewmodel/SettingsViewModel.java
```diff
@@ -0,0 +1,4 @@
+package gui.viewmodel;
+
+public class SettingsViewModel extends ViewModel{
+}
```
jenci2114, line 4: Do we need this class since it has no content?

valyippee: xander will be using it for ics exports

src/main/java/gui/GUIDriver.java
valyippee: edited!

